PR #6, opened 2017-12-08
Turn off default gc ordering of EVs in cis

jnuebler:
Set default ordering of EVs to sorting by eigenvalue (sort_by_gc_corr=False) in cis eigenvalue decompositions and added recommendation accordingly


golobor:
cleaned up some docstrings

PR #18, opened 2018-04-16
added color-range for saddleplot

sergpolly:
color range option added for saddle-cli: `--crange`.

Problem was, that for the default `crange=(-1,1)` (which was `vmin=-1`, `vmax=1`, internally), most of the saddleplots would look pale, lack contrast. This should fix it, by adding `--crange -0.5 0.5` for example.


sergpolly:
`--crange` in action, after all the colorer ticks adjustments:

`--crange -1 1`
xxx

`--crange -0.5 0.5`
zzz


`--crange -0.3 0.3`
yyy

Phlya:
Minor thing, but why not use `norm=LogNorm()` and have colorbar with log-spaced ticks with linear values? Just looks neater I think... Like this: https://matplotlib.org/examples/pylab_examples/hist2d_log_demo.html

sergpolly:
What's the idea again ? is the default `colorbar`-ticks supposed to look better, when `norm=LogNorm()`? or just overall color scale for heatmap would adjust better by default?

That's what I got, by tweaking https://github.com/mirnylab/cooltools/blob/cdd1163489e72b7b91e08231dff746935ac44c50/cooltools/saddle.py#L261 and removing every mention of `vmin,vmax`, including from here https://github.com/mirnylab/cooltools/blob/cdd1163489e72b7b91e08231dff746935ac44c50/cooltools/saddle.py#L327 :

rrr

I also tried `norm=LogNorm(vmin=0.3,vmax=3.0)` and manually setting `cb_ticks=[0.3,0.9,1.5,3.0]` :
ttt

sergpolly:
After playing with `colors.LogNorm` a bit , It feels like the only principal difference here is that we are getting linear-scaled values on the colorbar, while colors are log-scaled. 
Every other aspect of this would require just as much of manuall tweaking as the current implementation

sergpolly:
@golobor @nvictus is that something that we want to do now ?
Something like `--crange` of any implementation is needed for people , as the `vmin,vmax=-1,1` saddleplots looks pale.

Phlya:
No-no, keep the vmin and vmax! It's just an additional thing, not a substitute, I just noticed you were doing stuff with this anyway, so thought I'd mention this.
"It feels like the only principal difference here is that we are getting linear-scaled values on the colorbar, while colors are log-scaled." Yes, exactly, I think this is a better visualization practice (although I would also apply some formatter to get rid of the scientific notation, since here we would only have a very limit data range and not many orders of magnitude). Maybe it's just my opinion, so really feel free to dismiss this.
(But yeah, I have a feeling something with the ticks is a little broken in matplotlib currently, like in your first plot all the ones above the center are missing, but manually setting them is not so difficult)

nvictus:
Looks good. I agree with @Phlya about using `--vmin` and `--vmax` instead of `--crange`. Good to stick with matplotlib idioms if possible. I updated the PR. If it works, I'll merge.

PR #22, opened 2018-05-16
Eigdecomp cli draft

sergpolly:
`call_compartments` and `compute_phasing` added (drafted rather).

`call_compartments` isn't fully finished, but working in the current form.
`compute_phasing ` is of course feels a bit out of place in `cooltools` but is needed , kinda , badly. Only 'gc' phasing is implemented. I'm experimenting with the gene density - it seems like `bioframe.tools.frac_gene_coverage` does not work the way it is because of `bedtools.coverage` call - `bedtools` cmd and subcommand are defined downstream it seems, or maybe require unusual `import` ?! Should `compute_phasing` be 1 script or should there be separate  ones for `gc` and `gene_density ` ? 
This seems relevant: https://github.com/mirnylab/bioframe/issues/5


golobor:
this is awesome, thank you so much for fixing it!!

Re:phasing - I still wonder if that's the best way to name this procedure? That is, GC and gene density could be useful outside of eigenvector decomposition, but the suggested name may focus too much on that particular use of these tracks. Not a biggie, though, it's just great to have CLIs for these functions! @nvictus , what do you think?

Phlya:
Yeah, why is it called phasing?.. What is usually meant by phasing is splitting the genome into maternal/paternal genomes by SNPs and similar things?

golobor:
oh, true, "phasing" is not ideal, too! Indeed, genome phasing is an act of
diploid genome reconstruction. In our case, it's more like "flipping" +
"selection" (among the top eigenvectors). Maybe "conditioning"? @nvictus
usually has great ideas for naming :)

sergpolly:
I was not aware of `phasing`-naming issues, but my biggest concern is that this CLI-tool feels out of place here in `cooltools`.
And personally I'd prefer a binned-gc calculator and binned gene density calculator to be separate CLI tools, just to keep things simple, and as @golobor said , to avoid focusing too much on just one use case

gfudenberg:
I would call it orientEigenvector, possibly orientEigenvectorByGC

nvictus:
> Should compute_phasing be 1 script or should there be separate ones for gc and gene_density ?

I think it would be best if bioframe's functionality were exposed as a CLI through cooltools and keep bioframe strictly as library. I was thinking of having a command group like `cooltools genome <subcommand>` to handle various "binned genome" tasks, including statistics on fasta files.

re: the name "phasing". I have no good suggestions on nomenclature. Note that "phasing" here encompasses both orienting the eigenvectors and ranking them by agreement with the phasing track, though maybe those two aspects should be decoupled.

I'm actually a bit concerned about the need for extra options to specify both a "track file" and the appropriate column, because this scenario pops up often. I'm wondering if we could come up with a syntax that does both, similar to the "Cooler URI", which I borrowed from the [odo](http://odo.readthedocs.io/en/latest/uri.html) project. e.g. A simple text file like `gcContent.bed` could be assumed to be bedGraph by default (4th column is the value column), while `gcContent.bed::5` would instruct to use the 5th column, and `gcContent.bed::GC` would imply that a header exists and to use the column named `GC`. You could then even pass a URI-style path into a cooler's bintable: `mycooler.cool::bins/GC` or `h5py://mycooler.cool::bins/GC`. The same syntax could also be used for specifying outputs.

Of course, it would still be assumed that the first three columns are `chrom`, `start` and `end`.

sergpolly:
- `cooltools genome <subcommand>` - sounds good to me, we can even regroup that anytime ...
 - since `phasing` is an ambiguous nomenclature and makes people focus too much on just 1 aspect of `gc` , `gene_dens` -tracks, I think it makes sense to split the `compute_phasing` script in the PR into 2 (just `gc` for now, because of https://github.com/mirnylab/bioframe/issues/4)
 - I love the URI idea, we could standardize tab-separated files validation as @nvictus is proposing: https://github.com/mirnylab/cooltools/blob/70a04e4196d89821c89faec2692a17fb08e529e6/cooltools/cli/compute_saddle.py#L16
 - bit more documenting would be great, as I was super surprised by this URI thing in a `saddle cli` ...
 - I also like @Phlya -s proposal to store binned tracks inside of a cooler, that could be an option for `compute_gc` I guess. There is no ambiguity like in the chromosomes arm case, right @nvictus ? 
 - inside of the `call_compartment` script we could change `phasing` to `orient_and_rank` or something like that - I guess `call_compartment` is the only place where this issue pops up, especially if we split `compute_phasing` to `cooltools genome compute_gc` and `cooltools genome compute_gene_density` ...

nvictus:
Awesome! 

* Let's have a separate PR to start the `cooltools genome` command group.
* Let's also start a discussion on a convention for specifying sub-chromosomal regions as command line inputs (e.g. arms). That would go together with a `cooltools genome` command to return a list of chromosome arms for human assemblies.

> bit more documenting would be great, as I was super surprised by this URI thing in a saddle cli ...

Ha! I even forgot that I implemented that for saddle. I moved the validator to `cooltools.cli.util` for reuse and added a few changes to the PR. I think it can be elaborated on in the future to handle more use cases.

I renamed the phasing track input CLI option to `--orient-track` and made it optional. Still sounds a bit awkward, but `--orient-and-rank-track` would sounds worse. Thoughts, @gfudenberg?

Aside from more documentation, does this look good to merge?

sergpolly:
Looks awesome!

PR #31, opened 2018-06-12
Looping over window sizes in insulation score

itsameercat:
Updated develop branch with new insulation.py that loops over multiple window sizes.


nvictus:
Will merge into develop as-is. Make another PR for unit tests when you have the time.

PR #43, opened 2018-10-05
Scaling plot

DekkerBastiaan:
Added scaling script using pairix for fast random access.

### DIFF
--- a/cooltools/cli/__init__.py
+++ b/cooltools/cli/__init__.py
@@ -55,4 +55,5 @@ from . import (
     call_dots,
     call_compartments,
     genome,
+    scaling_script,
 )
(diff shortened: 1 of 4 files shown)


sergpolly:
We see so many commits here, likely because the `develop` branch is not up to date with `master` (needs to be updated anyways) - the scaling script itself is just a couple of commits at the very bottom.

golobor:
awesome, thank you!! I'll take a look

Phlya:
Just wanted to mention that output of `pairtools stats` contains all required information to make a scaling plot, so, for example, for distiller-analysed data a separate script might be redundant - but I'm travelling and haven't actually checked what this script is doing. And maybe this is useful anyway.

sergpolly:
pairs-related scripts have been moved to https://github.com/open2c/pairlib - so this no long applies

PR #67, opened 2019-03-20
Adding zoomArray from mirnylib.numutils

Phlya:
Hi, as we discussed, adding zoomArray function from mirnylib.numutils to cooltools.numutils. 

Seems like a side effect of just saving the files with Spyder cleaned up loads of trailing spaces, is that OK?


nvictus:
Deleting of trailing spaces is fine, actually a good thing. Could you also replace camelCasing of the function name and args with snake_casing for consistency?

Phlya:
Done
(still new to some aspects of git, excuse the extra commits...)

Phlya:
Also, as we discussed with Max, zoom should have order=1 to reduce interpolation artefacts, shall I just add this in the same PR?

nvictus:
I don't think I followed that conversation. If it's a simple addition, go ahead.

Phlya:
Basically, when I tried using the function with higher orders it introduced artefacts, but with order=1 it becomes a simple linear interpolation with much better results.

nvictus:
lgtm

Phlya:
Thanks!

PR #76, opened 2019-04-30
Add call-compartments to README

pkerpedjiev:
Provide an example of how to use `call-compartments` to calculate eigenvectors.


golobor:
thank you!

PR #80, opened 2019-06-05
cis and total calculation from coolers

gfudenberg:
with corrected main diagonal if it is not masked


golobor:
will edit in master :)

gfudenberg:
🕵

PR #89, opened 2019-06-28
Some changes to interp_nan

nvictus:


gfudenberg:
looks good! might be ~20% slower but that can be addressed in the future
image

also can add a warning if there isa large fraction of NaNs being interpolated-- highly nan regions can get crazy interpolations
image

PR #95, opened 2019-08-08
Small snipper fixes

agalitsyna:
Snipper fixes: NaN filling of sparse matrix, chromosomes arms expected retrieval and ExpectedSnipper for different regions added.


nvictus:
Thank you!

Issue #87, opened 2019-06-25
saddle plots for genomic bands - best way to specify options ?


sergpolly:
it seems like adding support for diagonal genomic bands in saddle plot cli is generally useful and interesting - it's essentially would do the same as @jnuebler COMPscore by genomic separation, but can be advantageous for shallow data, and for seeing nuances of compartmentalization - we recently used it in https://www.biorxiv.org/content/10.1101/678474v1 - here is an example:
![genomic_band_saddle](https://user-images.githubusercontent.com/6790270/60113705-5b7b0200-9740-11e9-9cd0-ff8a86fd4c51.png)

My question is what is the best way to add corresponding options to the saddle CLI script ?
diagonal regions #37 ? or should it be like `pairtools select`-type of thing:
```
pairtools select '(chrom1==chrom2) and (abs(pos1 - pos2) < 1e6)'
```
I personally like the general `pairtools`-style,  but maybe it is worth expanding the definition of a `region` - as there would probably be only "diagonal" and "rectangular" regions only  - maybe some crosses and combinations of the two, right ?

Phlya:
I am considering trying this, is there a way this can be currently achieved? Filling unwanted distances in expected with `np.nan`, perhaps?

sergpolly:
hey! @Phlya - yes, that's how we've done it - fill with `np.nan`-s whatever we didn't want to be counted in the saddle calculations
I uploaded that notebook here **https://github.com/dekkerlab/Abramo_2019**
this notebook is a mess, but the key-function, as far as I remember is `make_cis_obsexp_diagband_fetcher` - which you can plug in the "normal" saddle calculator from `cooltools` - the https://github.com/mirnylab/cooltools/blob/f67e626f1146be4f8448e61bff6fbfb179f4be70/cooltools/saddle.py#L205  - i.e. the `getmatrix` argument, as a replacement of this one: https://github.com/mirnylab/cooltools/blob/f67e626f1146be4f8448e61bff6fbfb179f4be70/cooltools/saddle.py#L94

sergpolly:
in this post I was hoping to get some input on how to best generalize such specification of genomic bands - sort of like `regions` argument in `make_saddle` but for genomic separations ...

sergpolly:
by @Phlya suggestion:
```
--min-distance
--max-distance
```

sergpolly:
[...]

PR #102, opened 2019-10-29
Saddle distance

Phlya:
Allow `--min-dist` and `--max-dist` arguments for `compute-saddle`.

Addresses https://github.com/mirnylab/cooltools/issues/54

Here is an example with some mouse ES cell data with mindist from 0 to 48 in 12 Mb increments, for 12 Mb bands (like in the Abramo paper). 

image

### DIFF
--- a/cooltools/saddle.py
+++ b/cooltools/saddle.py
@@ -190,7 +190,7 @@ def _accumulate(S, C, getmatrix, digitized, reg1, reg2, min_diag, max_diag,
     if reg1[0] == reg2[0]:
         for d in np.arange(-min_diag+1, min_diag):
             numutils.set_diag(matrix, np.nan, d)
-        if max_diag < np.inf:
+        if max_diag >= 0:
             for d in np.append(np.arange(-matrix.shape[0], -max_diag),
                                np.arange(max_diag+1, matrix.shape[0])):
                 numutils.set_diag(matrix, np.nan, d)
@@ -209,7 +209,7 @@ def _accumulate(S, C, getmatrix, digitized, reg1, reg2, min_diag, max_diag,
 
 
 def make_saddle(getmatrix, binedges, digitized, contact_type, regions=None,
-                min_diag=3, max_diag=np.inf, trim_outliers=False, verbose=False):
+                min_diag=3, max_diag=-1, trim_outliers=False, verbose=False):
     """
     Make a matrix of average interaction probabilities between genomic bin
     pairs as a function of a specified genomic track. The provided genomic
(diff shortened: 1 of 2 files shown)


cooltools/saddle.py
```diff
@@ -226,6 +232,10 @@ def make_saddle(getmatrix, binedges, digitized, contact_type, regions=None,
     regions : sequence of str or tuple, optional
         A list of genomic regions to use. Each can be a chromosome, a
         UCSC-style genomic region string or a tuple.
+    min_diag : int
```
nvictus: Here, too. Only applies when contact_type = cis.

nvictus:
lgtm!
@sergpolly, does this also resolve #87 ?

sergpolly:
@Phlya @nvictus - yes looks like this would take care of #87 as well ...
Thank you @Phlya !
i'll leave here the way we've done it - just for history:
```python
def make_cis_obsexp_diagband_fetcher(clr, expected, name, band_width_bins, inside=True):
    """
    Given a cooler object 'clr',
    cis-expected dataframe 'expected',
    and a column-name in cis-expected 'name',
    this function yields a function that
    returns slice of OBS/EXP for a given 
    chromosome.
     Custom one with an additional band selector ...
    """
    full_obsexp = clr.matrix().fetch(chrom) / \
                toeplitz(expected.loc[chrom][name].values)
    i,j = np.indices(full_obsexp.shape)
    i = i.flatten()
    j = j.flatten()
    # fill NaNs outside of the desired selection:
    nan_band_selector = (np.abs(i-j) > band_width_bins) \
                     if inside else (np.abs(i-j) <= band_width_bins)
    full_obsexp[i[nan_band_selector],j[nan_band_selector]] = np.nan
    return full_obsexp
```

I also wonder if `numutils.set_diag(matrix, np.nan, d)` for a lot of diagonals is slower than the modified `getmatrix` approach ...

sergpolly:
something like that to be tested:
```python
def getmatrix(clr, expected, name, min_dist, max_dist):
    """
     ...
    """
    OE = clr.matrix().fetch(chrom) / \
                toeplitz(expected.loc[chrom][name].values)
    i,j = np.indices(OE.shape)
    i = i.flatten()
    j = j.flatten()
    # fill NaNs wherever needed:
    dist_mask = np.abs(i-j) > min_dist
    dist_mask = dist_mask & np.abs(i-j) < max_dist
    OE[i[dist_mask], j[dist_mask]] = np.nan
    return OE
```
vs 
```
for d in diagonals_tobe_filled_with_nans:
    numutils.set_diag(matrix, np.nan, d)
```

nvictus:
Interesting.. then we can keep #87 open to consider alternative `getmatrix` implementations.

PR #103, opened 2019-10-30
Filter bad bins in track for saddle

Phlya:
Adds filtering of bad bins in the track based on Hi-C weights - removes those that were filtered in pre-balancing filtering.

Example before and after:
image


golobor:
awesome!! looks very clean to me. The code assumes that the cooler and the track table share bins; I can't remember how do we prepare these tracks from chip-seq?
Also, since you're on it, could you please also fix the example jypyter notebook around this line? While eigenvectors are most likely already nan-ed at bad bins, it be nice to have a line for masking (even if commented out) such that users would be aware of this operation.
image
Many thanks again!

Phlya:
Thanks!

Yeah, I realized this is an assumption - but I think it's safe to say that if bins are not aligned it's not really worth doing anything with thee data anyway? Problem could be when ChIP-seq bins e.g. start from 1 instead of 0, while actually correspond to the same locations... But then, we are also not guaranteed that all chromosomes from the cooler are present in the track, so we can't just assume they are all in the same order.

Something more complicated could be for each chromosome check they are the same size in both tables, and if they are - just filter based on the index, and if not - raise an error (slightly more strict test than the comparison for total number of rows in validation https://github.com/mirnylab/cooltools/blob/ececf9127a1425bd001327f8b9fd2719d4ce5629/cooltools/cli/compute_saddle.py#L260).

I, actually, should check whether this works as expected with missing chromosomes in the track...

golobor:
yeah, I'd say it's a subject for another PR, since it may involve a large-scale overhaul of our assumptions and extra data checks. For now, what you posted is good!

golobor:
make it 
image

:-P

looks good to me, thank you!

PR #105, opened 2019-11-01
Saddle scale

Phlya:
As per our discussion: saved saddledata is not transformed while plotting is done with LogNorm.


nvictus:
Awesome! I think this is ready to merge. One last thing (something we should start doing): could you add a short line to the changelog describing this change?

Phlya:
Hope this description is fine?

nvictus:
Looks great!

PR #109, opened 2019-11-05
Error for non-positive vmin or vmax in saddle

Phlya:
Should I add an error/explanation to the Python function too, or just in CLI is enough?


golobor:
I think having it in CLI would probably be enough?..

PR #117, opened 2019-12-05
Minor CLI update: cleanup, consistency, raw expected

sergpolly:
- all of the changes are only in CLI
 - added `--balanced/--no-balance` (same as in `cooler dump`) to the CLI; `no-balance` would proceed using `transforms={}` - i.e. only `count.sum` and `.avg` would be reported
 - define `transforms` and `.avg` columns (`.sum/n_valid`) outside of the `try`-clause where `diagsum` is calculated and thus made it common for `cis/trans`

some minor consistency changes `drop-diags`->`ignore-diags` just like in cooler and in other cooltools
`balancing-weight-name`-> `weight-name` for `call-dots` CLI - just for consistency
quickly tested CLI - ok


cooltools/cli/compute_expected.py
```diff
@@ -47,41 +47,35 @@
     "of a Hi-C map.",
     type=click.Choice(['cis', 'trans']),
     default='cis',
-    show_default=True,
-    )
+    show_default=True)
+@click.option(
+    "--balanced/--no-balance",
```
nvictus:
```suggestion
    "--balance/--no-balance",
```

sergpolly: more consistencies to be fixed later between `cooler` and `cooltools`, e.g. https://github.com/mirnylab/cooler/blob/843dadca5ef58e3b794dbaf23430082c9a634532/cooler/cli/dump.py#L273

cooltools/cli/compute_expected.py
```diff
@@ -91,52 +85,60 @@ def compute_expected(cool_path, nproc, chunksize, output, hdf, contact_type, wei
...
                     for (s1, s2), rec in records.items()],
                 columns=['chrom1', 'chrom2', 'n_valid',
                          'count.sum', 'balanced.sum'])
-            result['balanced.avg'] = result['balanced.sum'] / result['n_valid']
     finally:
         if nproc > 1:
             pool.close()
 
+    # calculate actual averages by dividing sum by n_valid:
+    result['count.avg'] = result['count.sum'] / result['n_valid']
```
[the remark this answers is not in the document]
sergpolly: related - just realized that this https://github.com/mirnylab/cooltools/blob/cb77cf219aa38286f418b0a15eb44128de295d4e/cooltools/expected.py#L775
prevents us from calculating `trans` expected for unbalanced data ...

I know using `n_total` alone overestimates denominator, but maybe we could allow that in case someone really needs that ?

application in mind: very low resolution data ~1-5Mb bins for shallow data, e.g. for looking for translocations and other Structural Variation in cancer cell lines etc - in this cases balancing might not be applicable to begin with ...
what do you think ?

PR #126, opened 2019-12-17
Update dotfinder.py

gspracklin:
Note sure if this is the most efficient solution but it solved the int to float issue for chrom positions
@sergpolly @nvictus 

Warning:
/home/spracklin/devel/cooltools/cooltools/dotfinder.py:1776: FutureWarning: 
Passing list-likes to .loc or [] with any missing label will raise
KeyError in the future, you can use .reindex() as an alternative.

Solution found in:
https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#deprecate-loc-reindex-listlike

### DIFF
--- a/cooltools/dotfinder.py
+++ b/cooltools/dotfinder.py
@@ -1773,7 +1773,7 @@ def clustering_step(
 
     # report only centroids with highest Observed:
     chrom_clust_group = df.groupby(["chrom1", "chrom2", "c_label"])
-    centroids = df.loc[chrom_clust_group[obs_raw_name].idxmax()]
+    centroids = df.loc[df.index.intersection(chrom_clust_group[obs_raw_name].idxmax())]
     return centroids
 
 
--- a/cooltools/dotfinder.py
+++ b/cooltools/dotfinder.py
@@ -1770,10 +1770,12 @@ def clustering_step(
     df = pd.merge(
         scores_df, pixel_clust_df, how="left", left_index=True, right_index=True
     )
-
+    #prevents scores_df categorical values (all chroms, including chrM)
+    df['chrom1'] = df['chrom1'].astype(str)
+    df['chrom2'] = df['chrom2'].astype(str)
     # report only centroids with highest Observed:
     chrom_clust_group = df.groupby(["chrom1", "chrom2", "c_label"])
-    centroids = df.loc[df.index.intersection(chrom_clust_group[obs_raw_name].idxmax())]
+    centroids = df.loc[chrom_clust_group[obs_raw_name].idxmax()]
     return centroids
 
 


sergpolly:
This must be it then ...
But ... the error itaself does not make sense to me still:
in your pandas doc link see this section:
> Selection with all keys found is unchanged.

that implies that if we are `.loc[]`-selecting a bunch of rows with existing indexes it should still work:
```
chrom_clust_group = df.groupby(["chrom1", "chrom2", "c_label"])
centroids = df.loc[chrom_clust_group["count"].idxmax()]
```
`idxmax` returns a subset of indices from `df` itself - it's a subset of indices that correspondond to the "brightest" pixels in each cluster of pre-filtered dots ... - this MUST work unless there is something "bad" (e.g. non-unique) with indices in the `scores_df` (i.e. `filtered_pixels_qvals` in the call dots CLI) 

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.DataFrameGroupBy.idxmax.html

this might be relevant https://stackoverflow.com/questions/32459325/python-pandas-dataframe-select-row-by-max-value-in-group

sergpolly:
this is to https://github.com/mirnylab/cooltools/issues/125

sergpolly:
@gspracklin could you post the content of `df` after the merge (with indices) (from inside of the CLI version that is failing):
```
df = pd.merge(  scores_df, pixel_clust_df, how="left", left_index=True, right_index=True )
```
along with whatever `idxmax` returns:
```
chrom_clust_group = df.groupby(["chrom1", "chrom2", "c_label"])
chrom_clust_group["count"].idxmax()
```

I just checked indices in `Unsync_10kb.bedpe.filteredpixelsqvals.txt`, which is supposedly `scores_df` that goes into `clustering_step`, - everything looks OK - unique looking indices running from 0 to whatever ...

 - `df` after `merge` should be of the same length as `scores_df` and have THE SAME indices
 - `idxmax` should return a subset of those OK-looking indices

gspracklin:
@sergpolly 

Debugged more with Nezar...

Looks like score_df has all chromosomes (including chrY and chrM) as categorical values. During the groupby step, all possible chrom combinations are made for the number of c_labels (see attached). 

seems related to this: 
https://github.com/pandas-dev/pandas/issues/17594 

[idxmax_count_groupby.txt](https://github.com/mirnylab/cooltools/files/3985633/idxmax_count_groupby.txt)

gspracklin:
This doesn't explain why I am specifically seeing this issue (pandas v0.25.3)

xinyangbing:
Hi guys,
Is this problem fixed? I am getting this error when I am trying to run call-dots on CLI.

Issue #120, opened 2019-12-06
extending CLI expected and expected.py module


sergpolly:
There are several things to be addressed:

1. `blocksum_pairwise` - would require new parameters , e.g. `weight_name` to allow different balacning names https://github.com/mirnylab/cooltools/blob/cb77cf219aa38286f418b0a15eb44128de295d4e/cooltools/expected.py#L771
2. allow `blocksum_pairwise` to divide sums by the `n_total` - without subtracting `n_bad` - maybe usefull for non-balanced data (low resolution, SV analysis stuff)
3. `diagsum` effectively divides by `n_total` , in case when `transforms={}` provided - i.e. when trying to calculated expected on the raw data
4. is it a good idea to be able to apply `blocksum_pairwise` to inter-arm regions? they are almost "flat"... ?
5. of course we'd need to introduce `--range` (and potentially `--range2`) to calculate intra-armal expected (which might be useful for dot calling btw)
6. turn `expected` into scaling plots/distance decay calculated from `cooler` (i.e. binned data)
7. what am I missing ?

sergpolly:
basically a subset of https://github.com/mirnylab/cooltools/issues/37

Phlya:
@sergpolly can you check if this can be closed? You addressed some things in the PR, and I _think_ the rest is covered in #37?

sergpolly:
sure ! yeah ! mostly addressed now

PR #127, opened 2019-12-18
updates to expected blocksum - weights, etc

sergpolly:
starting to address #120 :

for now about points 1 and 2 - related to the `blocksum`:
 - put `n_bad_block_elements` and `n_total_block_elements`(unchanged) outside of the `blocksum`
 - added `weight_name` parameter to `n_bad_block_elements`
 - `weight_name` could be `None` - yields 0 bad pixels per combination of support-regions in this case

todo:
 - update `blocksum_pairwise` to include `weight_name` parameter ...
 - update CLI `expected` correspondingly
 - planning to rewrite https://github.com/mirnylab/cooltools/blob/cb77cf219aa38286f418b0a15eb44128de295d4e/cooltools/expected.py#L401 using `n_bad_block_elements` and `n_total_block_elements` - should we do it or get rid of `trans_expected` altogether ? or maybe rewrite `trans_expected` using `blocksum_pairwise`


nvictus:
> should we do it or get rid of trans_expected altogether ? or maybe rewrite trans_expected using blocksum_pairwise

Yeah, `cis_expected` and `trans_expected` are deprecated and should either be removed or turned into aliases or higher-level versions of `diagsum` and `blocksum`.

cooltools/expected.py
```diff
@@ -417,77 +499,48 @@ def trans_expected(clr, chromosomes, chunksize=1000000, use_dask=False):
...
-
-    # returning a DataFrame with MultiIndex, that stores
-    # pairs of 'balanced.sum' and 'n_valid' values for each
-    # pair of chromosomes.
-    dtable = pd.merge(
-        trans_sum.to_frame(), trans_area.to_frame(), left_index=True, right_index=True
-    )
+
+    # turn chromosomes into supports:
+    chrom_supports = [ (chrom, 0, None) for chrom in chromosomes ]
```
sergpolly: @nvictus I'm using this feature of `lib.common.assign_supports`: https://github.com/mirnylab/cooltools/blob/9c8028f756865f8315096419b3d32bc7d6560bed/cooltools/lib/common.py#L34

the feature is that `(chrom,0,None)` implies the entire chromosome , not sure is `clr.chromsizes[chrom]` would be better instead of `None`

sergpolly:
was reimplementing `cis_expected` using `diagsum` and `diagsum_asymm` ...
found a discrepancy between current (1) `cis_expected` and (2) `diagsum_asymm`:
screenshot_1

the scatter is for `balanced.sum` for (1) and (2) for region: `[("chr7",10000000,20000000),],[("chr7",20000000,30000000),]`

corresponding `diagsum` -i.e. symmetric case looks fine - `np.allclose(...,equal_nan=True) -> True` ...

@nvictus we should probably investigate, right?

sergpolly:
a lot of it have been discussed here https://github.com/mirnylab/cooltools/issues/37 ...

sergpolly:
regarding the discrepancy - it seems to be coming down to the difference between `assign_supports` and `bg2slice_frame`:
`bg2slice`-thing does:
```python
(pixels["chrom1"] == chrom1) & (pixels["start1"] >= start1) & (pixels["end1"] < end1) &
(pixels["chrom2"] == chrom2) & (pixels["start2"] >= start2) & (pixels["end2"] < end2)
```
while `_diagsum_asymm`'s:
```python
pixels["support1"] = assign_supports(pixels, supports1, suffix="1")
pixels["support2"] = assign_supports(pixels, supports2, suffix="2")
```
does this:
```python
c = "chrom" + suffix
s = "start" + suffix
e = "end" + suffix
sel = (features[c] == region[0]) & (features[e] >= region[1])
sel &= features[s] < region[2]
```
for `suffix=1` and `2`, where `region` should be `(chrom,start,end)` of some sort ...

which is equivalent to:
```python
(pixels["chrom1"] == chrom1) & (pixels["end1"] >= start1) & (pixels["start1"] < end1) &
(pixels["chrom2"] == chrom2) & (pixels["end2"] >= start2) & (pixels["start2"] < end2)
```
i.e. `start` and `end` are swapped with respect to each other...

it seems to me that `bg2slice` from `cooltools.expected` does the right thing - and `assign_supports` is currently a little broken - could someone with a clear mind look and confirm that please @nvictus @golobor  ?

sergpolly:
indeed - swapping `s` and `e` inside this region-loop https://github.com/mirnylab/cooltools/blob/9c8028f756865f8315096419b3d32bc7d6560bed/cooltools/lib/common.py#L30 makes `diagsumm_asymm` and `cis_expected` generate the same result

there are still some discrepancies between what one get with `clr.matrix().fetch(supports1[0],supports2[0])` and either  `diagsumm_asymm`/`cis_expected` - `matrx.fetch` show more than `expected` is using to calculate:
screenshot_2

potentially a discrepancy between `make_diag_tables` and the way we select `pixels` for calculating sums ...

sergpolly:
yeah the latter discrepancy comes down to the fact that selecting pixels, like so:
```python
    out = bg2[
        (bg2["chrom1"] == chrom1)
        & (bg2["start1"] >= start1)
        & (bg2["end1"] < end1)
        & (bg2["chrom2"] == chrom2)
        & (bg2["start2"] >= start2)
        & (bg2["end2"] < end2)
    ]
```
in a lot of cases - yields 1 less column and 1 less row worth of pixes than:
```python
clr.matrix(balance=DOESNOTMATTER).fetch((chrom1,start1,end1),(chrom2,start2,end2))
```

sergpolly:
we agreed with @nvictus that retiring `trans_expected` and `cis_expected` in a very-graceful way is not really worth it - so we'll do following:
 - yield a `warning` of deprecation for both `cis/trans_expected`
 - keep `cis_expected` as is (because of the discrepancy between `bg2slice` and `assign_suports`)
 - BTW - `bg2slice` is likely doing the wrong thing - at least it does not follow the same logic as `clr.matrix().fetch()` - so that's bad and asymmetric expected would be a bit off in case of `cis_expected` ...
 - as far as `expected` CLI goes - we want to consider 3 cases:
    1. your `cooler` is balanced (got `weights` of some sort) and you want to work on balanced values 
    2. your `cooler` is balanced (got `weights`) but you want to work on raw counts, but still use `weigths` for masking `bad`-regions/bins
    3. your `cooler` is not balanced, impossible to balance, you don't want to, don't give an f , etc - and you may or may not want to mask some bins/regions  - but you still want your `expected` to be calculated ! (applicable for sparse data, contigs etc ....)

that's about it
to be continued

cooltools/cli/compute_expected.py
```diff
@@ -51,14 +51,23 @@
...
     "-t",
-    help="compute expected for cis or trans region of a Hi-C map.",
+    help="compute expected for cis or trans region of a Hi-C map."
+         "Ignored when genomic-regions are provided",
     type=click.Choice(["cis", "trans"]),
     default="cis",
     show_default=True,
 )
+@click.option(
+    "--genomic-regions",
```
sergpolly: this is just a placeholder for genomic regions as a bed/bedpe file ...
but I actually don't like it - i think it would make code really complicated - esp. inferring type of expected for every region or a pair of regions ...
the original idea in #127 
> of course we'd need to introduce --range (and potentially --range2) to calculate intra-armal expected (which might be useful for dot calling btw)

seems much more feasible flexible and nice ...

Although it would require ~40 CLI calls to calculate cis-expected per-arm ...

Not sure how to strike a balance between flexibility, practicality and simplicity of implementation
@nvictus @golobor  what do you think ?

cooltools/cli/compute_expected.py
```diff
@@ -70,6 +79,14 @@
     default="weight",
     show_default=True,
 )
+@click.option(
```
sergpolly: this is taken from `cooler.balance` , not sure if `blacklist` is the best keyword here

cooltools/cli/compute_expected.py
```diff
@@ -104,6 +143,8 @@ def compute_expected(
         weight2 = weight_name + "2"
         transforms = {"balanced": lambda p: p["count"] * p[weight1] * p[weight2]}
     else:
+        # no masking bad bins of any kind, when balancing is not applied
+        weight_name = None
```
sergpolly: when `--no-balance` - there will be no masking until `blacklist` is ready
how else can we do it ?

cooltools/expected.py
```diff
@@ -221,6 +201,97 @@ def count_all_pixels_per_diag(n):
     return np.arange(n, 0, -1)
 
 
+def count_all_pixels_per_block(clr, supports):
```
sergpolly: `count_all_pixels_per_block`
`count_all_pixels_per_diag`

`count_bad_pixels_per_block`
`count_bad_pixels_per_diag`

such a naming might be misleading, because it hints a unified API, which is not - `per_diag` ones are unaware of `cooler/supports` - they deal with matrix dimensions, while `per_block` are `clr/supports` based

but consistent naming is kind of nice otherwise ...

cooltools/expected.py
```diff
@@ -311,6 +382,37 @@ def cis_expected(
...
 
     """
+    warnings.warn(
+        "`cooltools.expected.cis_expected()` is deprecated in 0.3.2, will be removed subsequently. "
+        "Use `cooltools.expected.diagsum()` and `cooltools.expected.diagsum_asymm()` instead.",
+        category=FutureWarning,
+        stacklevel=2,
+    )
+
+    def _bg2slice_frame(bg2, region1, region2):
```
sergpolly: as we know with @nvictus  `cis_expected` doesn't do a good job at slicing pixels - at least it does it differently than `cooler.matrix().fetch()` - it's all because of `_bg2slice_frame` - which is why I made it internal function - and basically prepared the whole `cis_expected` for removal with the next release

cooltools/expected.py
```diff
@@ -417,101 +519,124 @@ def trans_expected(clr, chromosomes, chunksize=1000000, use_dask=False):
...
-            for chrom in chromosomes
-        ]
-        pairblock_list = []
-        for i in range(len(x)):
-            for j in range(i + 1, len(x)):
-                # appending to the list of tuples
-                pairblock_list.append((chromosomes[i], chromosomes[j], x[i] * x[j]))
-        return pd.DataFrame(pairblock_list, columns=["chrom1", "chrom2", "n_bad"])
+    warnings.warn(
+        "`cooltools.expected.trans_expected()` is deprecated in 0.3.2, will be removed subsequently. "
```
sergpolly: `trans_expected` is reimplemented using `blocksum_pairwise` and ready for retirement/removal

cooltools/expected.py
```diff
@@ -417,101 +519,124 @@ def trans_expected(clr, chromosomes, chunksize=1000000, use_dask=False):
...
-    return dtable
+    trans_df["balanced.avg"] = trans_df["balanced.sum"] / trans_df["n_valid"]
+    return trans_df
 
 
 ###################
 
 
-def make_diag_tables(clr, supports):
+def make_diag_tables(clr, supports, weight_name="weight", bad_bins=None):
```
sergpolly: `weight_name="weight", bad_bins=None` added here and below
`bad_bins` is just a placeholder for now

sergpolly:
@nvictus I think i addressed most of what we've discussed  - so this PR should be fine

although i'm not sure about the way to specify custom regions in  CLI-expected - it is definitely the time to make people start calculating per-arm expected and such,  but first, they need an appropriate tool

cooltools/cli/compute_expected.py
nvictus: Tentative resolution: Call it `--regions`. It will take a BED4-style file (chrom, start, end, name) so that each region gets a unique name. Keep as placeholder for now.

nvictus:
Placeholders to implement:
* `blacklist`
* `bad_bins`

PR #134, opened 2020-01-26
fixing failing docs build

sergpolly:


sergpolly:
@nvictus that box - to allow maintainers to added is just gone ... before submitting PR, after submitting PR - it's not there
 - maybe because I'm PR-ing from out master
 - maybe because i'm PR-ing as an organization
no idea - but I remember that box before - I kept it checked all the time

I tried to add you as a collaborator to `dekkerlab/cooltools` fork - maybe that would help ?!
I could also try to PR this from some other place ...

### About the failing docs:
`sphinx_click` says we have to have a package and all dependencies installed: https://github.com/click-contrib/sphinx-click#usage and it seems like `autodoc_mock_imports` is not helping at all ...

so in this PR I added whatever was needed in `requirements` and updated `Makefile` with the trick from the last answer here https://stackoverflow.com/questions/10057476/how-do-i-use-sphinx-with-cython - also @golobor seem to have much of the `pairtools` requirements in the `requirement-doc` 

I don't know how to avoid recreating the whole environment and not really sure why `sphinx-click` would require that

sergpolly:
actually i just found https://readthedocs.org/projects/cooltools/builds/10328423/ - and it seems like everyhting is failing because of `pip` again ?!

does the order in wich  `pip!=20.0`  appears matter ?

Phlya:
If mocking isn't working properly, try this
https://github.com/alex-rudakov/sphinx-argparse/issues/87#issuecomment-576195894

nvictus:
Thanks for investigating, @sergpolly. I managed to get it working. 

The cached environment on readthedocs was initializing with the broken pip version, so I had to [wipe it](https://github.com/readthedocs/readthedocs.org/issues/6554). I also had to add `click` to the docs requirements. No need for cython, thankfully.

sergpolly:
Nezar, @nvictus - but the command line reference is empty  - check generated docs

Sphinx_click is still not happy! It either need a full environment, or @Phlya solution - which @nvictus uses for cooler

In other words, when I build this docs in FULL working environment - I get that gorgeous cli reference - all the commands, all of it ...

Issue #139, opened 2020-02-25
trans expected calculation might be wrong - # of NaNs underestimated


sergpolly:
we have been noting for a while now - that trans-saddles look a bit too "redish" - i.e. almost every pixels seems to be more than expected ...

making saddles only for different groups of chromosomes makes this issue just pop: take only large chromosomes (human) chr1-chr10 - things look "normal", taking only small ones - saddles just look RED... - not good we though...

i.e. seems like we are underestimating our trans expected for certain pair of chroms by a lot.
reason might be that we are underestimating number of bad bins (the area of NaNs in a trans-heatmap).

This is how it is right now: https://github.com/mirnylab/cooltools/blob/702d24c51feb8d1c87812ec082968fc1414c8060/cooltools/expected.py#L290

which is nothing but `Number_of_NaNs_chrA*Number_of_NaNs_chrB` - which is you can see on a figure below could be a HUGE underestimation, as it corresponds only to the part of "grey"  - stroked in a black box, but should be : 
```
Number_of_NaNs_chrA*Length_chrB + Number_of_NaNs_chrB*Length_chrA - Number_of_NaNs_chrA*Number_of_NaNs_chrB
```

![Drawing](https://user-images.githubusercontent.com/6790270/75205945-2dbce180-5743-11ea-8154-33c854f4e7cc.png)

golobor:
great find!! did you guys just fix it here: https://github.com/mirnylab/cooltools/pull/140 ?

sergpolly:
@golobor yes #140 is all about it. But we wanted to do a little educational moment out of it - we'll fix it and merge it later today

PR #140, opened 2020-02-25
Fix bad pixel counting in trans

nvictus:


cooltools/expected.py
```diff
@@ -266,28 +266,41 @@ def count_bad_pixels_per_block(clr, supports, weight_name="weight", bad_bins=Non
         raise NotImplementedError("providing external list \
             of bad bins is not implemented.")
 
+    # Get the total number of bins per region
+    n_tot = [np.diff(clr.extent(region)) for region in regions]
```
sergpolly: 1. the typo here `regions` should be `supports` ...
2. `np.diff` returns `ndarray` we should extract a scalar from it `np.diff().item()` is the new way:
```
DeprecationWarning: np.asscalar(a) is deprecated since NumPy v1.16, use a.item() instead
```

sergpolly: i PRed that to `nvictus/master`

sergpolly:
here is the same thing with the # of bad bins calculated in 2 different ways `np.sum(np.isnan(...))` , vs @nvictus PRed solution - no difference:
bad_bins_good

nvictus:
Fixes #139

PR #147, opened 2020-04-08
Set default zoom order to 1

Phlya:
Somehow default zoom order was set to 3 instead of a much better default of 1! Goes against the docstring too.


Phlya:
Just as a curiosity, it's critical for nan-containing inputs, such as a matrix made from default expected with NaN'ed first two diagonals - https://github.com/Phlya/coolpuppy/issues/18
Screenshot from 2020-04-08 11-49-46

PR #148, opened 2020-04-14
Expected refactoring

golobor:
- [x] rename supports → regions
- [ ] assign_supports:
    - [ ] Create a function that takes two sets of intervals, where the second is non-overlapping and sorted, and assigns the span of each set 1 interval in set 2 intervals. Make sure that it works with different resolutions, where spans of individual pixels may overflow chromosomal arms. Should end up in bioframe.
    - [ ] rewrite assign_supports as a generic array operation on binspans?.. I.e. overlap a region and a point, but w/o chromosomes. 
    - [ ] modify diagsums to use these new functions
- [ ] bioframe.tools.split_chromosomes 
    - [x] if centromere is missing, assume the split point at 0
    - [ ] Allow BED inputs
    - [ ] multiple split points per chromosome 
- [ ] clean up diag_assym 
    - [ ] what’s this regions1, regions2?..
- [ ] autogenerate transforms from weights? 
- [ ] in bioframe, set up an easy way to generate and retrieve chromosomal arms
- [ ] finalize the format of the expected table 
- [ ] store data inside coolers

### DIFF
--- a/cooltools/expected.py
+++ b/cooltools/expected.py
@@ -266,28 +266,44 @@ def count_bad_pixels_per_block(clr, supports, weight_name="weight", bad_bins=Non
         raise NotImplementedError("providing external list \
             of bad bins is not implemented.")
 
+    # Get the total number of bins per region
+    n_tot = []
+    for region in supports:
+        lo, hi = clr.extent(region)
+        n_tot.append(hi - lo)
+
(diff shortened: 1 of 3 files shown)


golobor:
@nvictus , @sergpolly 
this is the PR with my suggestions for expected refactoring. Not ready yet!

PR #154, opened 2020-05-05
Synchronize cooltools 0.3 with bioframe 0.0

nvictus:
re: https://github.com/mirnylab/bioframe/issues/27


itsameerkat:
Yeah makes sense. Seems like too many things are broken between the two versions

nvictus:
Tests do pass now. For some reason the travis badge isn't updating.

PR #160, opened 2020-05-07
fixed saddle code to accept bedslice

mimakaev:


itsameerkat:
So to deal with all the broken dependancies after recent pushes, @nvictus  pinned cooltools dependency on bioframe to <0.1 (see #154 ). This probably didn't fix the saddleplot code but in this version of bioframe, the function `bedslice` is called [slice_bedframe](https://github.com/mirnylab/bioframe/blob/8153e5842ba8f68340f231346124acbf6876b5b5/bioframe/core.py#L214) which is a wrapper around `bisect_bedframe` (There also exists another function named [bedslice](https://github.com/mirnylab/bioframe/blob/8153e5842ba8f68340f231346124acbf6876b5b5/bioframe/core.py#L156) in this version that essentially does the same thing but it accepts a groupby object instead of the dataframe). I think we need to be consistent so we should choose one of them and use it for all of cooltools. But there is another problems here. `bisect_bedframe` does not work. The following line will only work if bedf only has one chromosome.: https://github.com/mirnylab/bioframe/blob/8153e5842ba8f68340f231346124acbf6876b5b5/bioframe/core.py#L210

So question to you: should we require all of cooltools to use bedslice or slice_bedframe? If the latter, then we should fix it.

mimakaev:
I see. This is not master, but the new branch, which will depend on the new bioframe. 

I was more hoping for an immediate patch to make this code work, so I can keep testing new bioframe in my production code. If you know the correct function to replace bedslice with, and it's quick to implement, maybe just commit the temporary fix now, and we can decide on functions later...

itsameerkat:
Yeah okay makes sense

PR #167, opened 2020-06-24
make travis testing great again

sergpolly:
there were conflicts between our conda `environment.yml` and pip `requirements.txt` that lead to build failure
`numba`-s from conda and pip were fighting over the `llvmlite` with each other - it was brutal !

I removed a bunch of conflicting stuff from conda's `environment` because we're using only for travis, and gave priority to pip's `requirements.txt`  cause that would affect `pip install cooltools` I believe ...

anyways - travis seems to be working now !
I consider that as a quick patch - that's why a PR straight to mirny-`master` ...


sergpolly:
just checked - we DO use `numba` in `numutils`.
I kept `pytables` in `environment.yml`, because ... dunno - no reason - but I'm aware that it pulls some kind of `numpy` and then this `numpy` is actually removed by the `pip` ... - so not sure if `pytables` belong in `requirements.txt` ...

wait, `requirements.txt`  actually have a`tables` as a requirement - and it seems to be `PyTables` themselves - isn'it https://pypi.org/project/tables/ ?

so perhaps `pytables` should be gone from `environment.yml` as well ?

nvictus:
> so perhaps pytables should be gone from environment.yml as well ?

Yeah, the pip package is just called tables.

I recommend keeping only `numpy`, `cython` and `h5py` as conda dependencies. Personally, I create all my environments with those (and `ipython`) and install everything else with pip. It's been working quite smoothly for me for quite some time.

sergpolly:
@nvictus , so then we should duplicate `numpy` both in `environemnt.yml`  and `requirements.txt` , right ? same for `cython` then

we'd need to keep ALL of the requirements in the `requirements.txt` for the sake of `pip install` and `python setup.py install` - right ?

nvictus:
Yeah, and we could even get rid of `environment.yml` completely and write the necessary commands in the travis config. We're only using it for the CI environment.

sergpolly:
I've managed to fix these 4 lines, finally, not without some stupidity, but nonetheless ...
@nvictus how does that look ?

.travis.yml
```diff
@@ -16,9 +16,10 @@ before_install:
   # Useful for debugging any issues with conda
   - conda info -a
 install:
-  # Create test environment and install deps
-  - conda env create -n test-environment python=$TRAVIS_PYTHON_VERSION -f environment.yml
-  - source activate test-environment
+  # Create test environment and install deps, prefer defaults, use bioconda when needed
+  - conda create -n testenv -c defaults -c bioconda python=$TRAVIS_PYTHON_VERSION numpy cython h5py
```
nvictus, line 20:
```suggestion
  - conda create -n testenv python=$TRAVIS_PYTHON_VERSION numpy cython h5py
```

sergpolly: understood... there is nothing `bioconda`-related among `numpy,cython,h5py` ...

nvictus:
bellisimo!

sergpolly:
change the comment then ...refarding the channels ?

nvictus:
> change the comment then ...refarding the channels ?

okay

sergpolly:
oh, I guess you can't once review is over...
whatever, next time

PR #178, opened 2020-08-13
Allow --output in diamond-insulation

Phlya:
Addresses https://github.com/mirnylab/cooltools/issues/175

I didn't touch the `genome` tools though, should I?


sergpolly:
this seem innocent to me - i.e. no reason not to accept - keep the question regarding `genome` in the issue for now, i guess...

PR #186, opened 2020-09-30
add expected test, make saddle tests pass

sergpolly:
keep cleaning after the expected update ...

now just resolved the compatibility between current output of cli expected and whatever compute-saddle wants it to be (added `--regions` option to saddle, but it's not 100% functional yet)

also added a simple cli compute-expected test...

tests should pass now

### DIFF
--- a/cooltools/cli/compute_saddle.py
+++ b/cooltools/cli/compute_saddle.py
@@ -7,10 +7,12 @@ import sys
 import pandas as pd
 import numpy as np
 import cooler
+from bioframe import parse_regions
 from .. import saddle
 
 import click
 from .util import validate_csv
+from . import util
 from . import cli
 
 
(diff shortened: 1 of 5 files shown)


sergpolly:
oh no - wtheck ?
```
Processing dependencies for cooltools==0.4.0.dev0
error: cytoolz 0.11.0 is installed but cytoolz<0.11 is required by {'cooler'}
```

nvictus:
I had to introduce `cytools<0.11` for Python 2.7 in the last 0.8 release. I'll remove the restriction in cooler 0.9.

sergpolly:
> I had to introduce `cytools<0.11` for Python 2.7 in the last 0.8 release. I'll remove the restriction in cooler 0.9.

Oops - so - was it bad that I've removed cytoolz from cooltools requirements ? Or it is fine - since cooler pulls it anyway ?

nvictus:
It's fine if it's causing conflicts like that. We can re-introduce it later.

PR #189, opened 2020-10-05
fix trans saddles a bit, add trans tests

sergpolly:
saddles were broken for `--contact-type trans` - and this PR is a quick fix for that
also added a couple of trans related tests - compartment and saddle CLI , in order to make sure trans-business isn't neglected going forward - WE NEED MOAR TESTS !

also updated some docstrings -because it took me awhile to understand what's written in there ... and some vriable names as well for the same reason - less confusion , better maintenance !

### DIFF
--- a/cooltools/eigdecomp.py
+++ b/cooltools/eigdecomp.py
@@ -266,7 +266,6 @@ def trans_eig(
     A[:, is_bad_bin] = 0
 
     # Fake cis and re-balance
-    A = numutils.iterative_correction_symmetric(A)[0]
     A = _fake_cis(A, ~is_trans)
     A = numutils.iterative_correction_symmetric(A)[0]
     A = _fake_cis(A, ~is_trans)
--- a/tests/test_compartments_saddle.py
+++ b/tests/test_compartments_saddle.py
@@ -149,7 +149,7 @@ def test_trans_saddle_cli(request, tmpdir):
             "--n-bins", "30",
             "--scale", "log",
             in_cool,
-            f"{out_eig_prefix}.trans.vecs.tsv::E3", # 3rd EV stores checkerboard
+            f"{out_eig_prefix}.trans.vecs.tsv::E3",  # 3rd EV stores checkerboard
             out_expected
         ]
     )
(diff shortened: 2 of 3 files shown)


agalitsyna (approved):
Tested on several datasets and seems to work. Noticed that "cmap" and "verbose" parameters are not used in CLI. Might be easily fixed, though. 

Also, notices that trans works much slower than cis saddle plot construction. Can there be another reason except for a larger number of regions for the averaging?

golobor (approved):

golobor:
It looks like we now need an official obs/exp fetcher, which could be used both here and in snipping

tests/test_compartments_saddle.py
```diff
@@ -106,7 +106,7 @@ def test_trans_compartment_cli(request, tmpdir):
     test_trans_eigs = pd.read_table(out_eig_prefix + ".trans.vecs.tsv", sep="\t")
     r = np.abs(
         np.corrcoef(
-            test_trans_eigs.E3.values, np.sin(test_trans_eigs.start * 2 * np.pi / 500)
+            test_trans_eigs.E1.values, np.sin(test_trans_eigs.start * 2 * np.pi / 500)
```
sergpolly, line 109: with the fix in `eigdecomp` - `E1` reflects "checkerboard" again

cooltools/eigdecomp.py
```diff
@@ -266,7 +266,6 @@ def trans_eig(
     A[:, is_bad_bin] = 0
 
     # Fake cis and re-balance
-    A = numutils.iterative_correction_symmetric(A)[0]
```
sergpolly, line 269: this looks unnecessary , even more - it introduces chrom-sized biases in trans map with faked cis data ...

re-balancing should happen only after the cis was faked

sergpolly: in the end it was problematic only for weird cases like 2-chromosomes, and or several chromosomes of very different sizes - not an issue in a "typical" scenarios .

Also , when it fails - it still preserves the checkerboard signal, however it can "place" it in EV2/3... instead

PR #192, opened 2020-10-13
Add logbin-expected CLI

Phlya:
As discussed with @golobor, this adds a small CLI tool wrapping new functions by @mimakaev to logbin expected and make derivatives. I exposed most of arguments, but some (e.g. changing the default smoothing, or custom spread functions) are not possible to set through CLI at the moment (although some limited improvement is possible, at least change the level of Gaussian smoothing, for example).

Let me know what you think! I haven't tested all the options yet, but it works.


sergpolly:
sorry it took me so long - but I've finally got a handle on this ...
I managed to update your PR - I guess this is only possible because you set the permissions this way on your fork - the right way i mean
I think i'll add some stupid validations to the expected csv/tsv input - just like we did in other CLI tools (most of them at least) - maybe add some tests ?!
Should there be a good logbinning example of sorts ?! Something based on existing test files - I'll try to check myself but if you have some ideas - let me know !

Phlya:
Cool thank you! It wasn't that long! I think it's normal that you can update a PR? I didn't do anything to any permissions.

Validations make sense. I'm not sure how tests are written for CLI tools, do we have any?..

Re example: I think we should add it to the example notebook for Ps we've been working on. So it has both a Python and a CLI version.

Phlya:
But if you remember why I asked you to review, is to check that all the features are in the right branches: you seemed to think logbinning stuff has been merged into master, while this PR is aimed at develop.

sergpolly:
Regarding branches - it is all sorted out now , because Nezar pushed latest
changes from master to develop , so now develop is up to date and aligned
with the master

sergpolly:
Logbinning is both in develop and master now...

Phlya:
OK great, thanks!

cooltools/cli/logbin_expected.py
```diff
@@ -0,0 +1,180 @@
+import pandas as pd
+import numpy as np
+from functools import partial
```
sergpolly, line 3: I added a bunch of stuff to align with other cli-s we have (compute-saddle to be more specific)

cooltools/cli/logbin_expected.py
```diff
@@ -0,0 +1,180 @@
...
+
+    EXPECTED_PATH : The paths to a .tsv file with output of compute_expected.
+    Must include a header. Use the '::' syntax to specify a summary column name.
+    
+    OUTPUT_PREFIX: Output file name prefix to store the logbinned expected
+    (prefix.log.tsv) and derivative (prefix.der.tsv) in the tsv format."
+    """
+
+    # unpack expected path and name as generated by click's callback to validate_csv:
+    expected_path, exp_summary_name = expected_path
```
sergpolly, line 128: some validations of expected ...

cooltools/cli/logbin_expected.py
```diff
@@ -0,0 +1,180 @@
...
+        usecols=expected_columns,
+        dtype=expected_dtype,
+        comment=None,
+        sep='\t',
+        verbose=False,
+    )
+
+    lb_cvd, lb_slopes, lb_distbins = expected.logbin_expected(
+        cvd,
+        exp_summary_name=exp_summary_name,
```
sergpolly, line 153: makes sure other names of diagonal summaries are allowed ...
one example could be "count.sum" to calculate Pc from raw but binned data (could be useful) i guess ...

cooltools/expected.py
```diff
@@ -922,6 +922,7 @@ def blocksum_asymm(
 
 def logbin_expected(
     exp,
+    exp_summary_name="balanced.sum",
```
sergpolly, line 925: propagate custom summary column name to the logbin expected function itself

tests/test_expected.py
```diff
@@ -303,3 +303,36 @@ def test_trans_expected_regions_cli(request, tmpdir):
             # atol=0,
             equal_nan=True,
         )
+
+
+def test_logbin_expected_cli(request, tmpdir):
```
sergpolly, line 308: added a stupid logbin expected test - just to make sure the new CLI runs, and that it generates output - output contents are not checked

sergpolly (approved):
Should be good imho

Phlya:
Nice, thank you! I was going to allow a different column name after you had a look at this, thanks for doing it!

sergpolly:
let's accept it then ...

PR #172, opened 2020-07-07
cool_cis_eig: added bad_bins and docstring

sergpolly:
added `bad_bins` for quick masking (e.g. small translocations and such), added docstring, changed a couple of variable names - to make more readable

@mimakaev this should be small and innocent, could you check please ?

PR #194, opened 2020-10-21
Develop to master

sergpolly:
this would bring Ilya's CLI logbin-expected to master, and some minor changes to logbin-expected API to accomodate the CLI

not sure why diamond-insulation's `--output` addition showed up here, as it is already in `master`


sergpolly:
@nvictus i don't know why github thinks that we need to merge 94 commits ... - these are commits that we've merged earlier to master ... Is it because we've "Squashed" them and so those 94 aren't in the "master"'s history ?

it looks bizzare - if it keep coming back, we should probably branch a new develop from the current master ...
Should we just squash and merge anyways ? There are no groundbreaking differences between develop and master at this point

sergpolly:
looks like the same exact situation https://stackoverflow.com/questions/49200667/github-squash-and-merge-subsequent-pull-request-showing-all-previous-changes
apparently we shouldn't have been squashing-merging from develop to master ...squash-merge is more for 1-time feature branches

what should we do now ? squash and merge one last time and then "restart" the develop branch ?

sergpolly:
I'm also afraid of a situation like we had in here https://github.com/open2c/cooltools/pull/172 - when some changes got overwritten, or undone - I really didn't understand that one

But yeah, what can we do - just squash for the last time and then check everything carefully

mimakaev:
Yeah. I'm puzzled by the general approach here.  Is it best to create a new branch for every change, and if you ended up making too many commits / back-and-forth changes, then squashing it?

Issue #94, opened 2019-08-02
turn off chromosome name filtering in binnify command


nvictus:
use `read_chromsizes(..., all_names=True)`

agalitsyna:
Solved in recent pileup-update: https://github.com/open2c/cooltools/blob/pileup-update/cooltools/cli/genome.py#L23

PR #198, opened 2020-11-02
Pileup update and innocent small fixes

agalitsyna:
1. Test files download API created. It allows storing a list of data in the repo that will be downloaded from external resources upon request from user.  
2. NaN supports are now allowed in pileup with a warning, not a crash. 
3. Random sample is parallel now.
4. Fix of binnify issue with non-typical chromosome names. (Solved github issue: https://github.com/open2c/cooltools/issues/94)
5. metavariable TRACK_PATH for call-compartments added.
6. color map of saddle now propagates to the plot (and is not get lost on its way)


cooltools/cli/genome.py
```diff
@@ -23,11 +23,16 @@ def fetch_chromsizes(db):
...
+    help="Parse all chromosome names from file, not only default r\"^chr[0-9]+$\", r\"^chr[XY]$\", r\"^chrM$\". ",
+    is_flag=True
+)
+def binnify(chromsizes_path, binsize, all_names):
     import bioframe
 
-    chromsizes = bioframe.read_chromsizes(chromsizes_path)
-    bins = bioframe.tools.binnify(chromsizes, binsize)
+    chromsizes = bioframe.read_chromsizes(chromsizes_path, filter_chroms=not(all_names))
+    bins = bioframe.genomeops.binnify(chromsizes, binsize)
```
nvictus, line 35:
```suggestion
    bins = bioframe.binnify(chromsizes, binsize)
```

nvictus (commented):
Looks good to me. My only comment would be to consider using a JSON data manifest if the `external_test_files.tsv` gets too unwieldy, but it's fine for now.

cooltools/lib/common.py
```diff
@@ -67,3 +72,140 @@ def assign_regions_to_bins(bin_ids, regions_span):
...
+
+    Notes
+    -------
+    Designed following seaborn download test data examples:
+    https://github.com/mwaskom/seaborn/blob/4dd57d63e23e4628d8b7f41cdefdf64f7fefcf74/seaborn/utils.py#L427
+
+    """
+
+    if data_dir is None:
+        data_dir = os.path.abspath(os.path.join(os.path.dirname( __file__ ), '..', '..', 'tests/data/external/'))
```
nvictus, line 166: Oh, I just realized. This may not work as expected. The `tests` folder does not "ship" with the package when it is installed the usual way. I think the example data should go somewhere within the package structure, like `cooltools/datasets`.

cooltools/lib/common.py
```diff
@@ -67,3 +72,140 @@ def assign_regions_to_bins(bin_ids, regions_span):
     region_ids[mask_assigned] = regions_span["name"][bin_reg_idx_lo[mask_assigned]]
 
     return region_ids
+
+
+def download_test_data(name="all", cache=True, data_dir=None):
```
nvictus, line 77: Maybe call it "example" data instead of "test" data? or maybe just `download_dataset`?

cooltools/lib/common.py
nvictus: As in sklearn: https://github.com/scikit-learn/scikit-learn/tree/master/sklearn/datasets

agalitsyna: Good catch! Moved it to `cooltools/datasets`.

cooltools/lib/common.py
```diff
@@ -67,3 +72,140 @@ def assign_regions_to_bins(bin_ids, regions_span):
...
+        downloaded = download_file(url, file_path)
+
+    return downloaded
+
+
+def get_data_dir(data_dir=None):
+    """
+    Returns a path to cache directory for example datasets.
+
+    This directory is then used by :func:`download_test_data`.
```
mimakaev, line 144: download_test_data was renamed to download_data

cooltools/lib/common.py
```diff
@@ -67,3 +72,140 @@ def assign_regions_to_bins(bin_ids, regions_span):
...
+
+    By default, this function uses the location of cooltools as the home folder
+    and stores the files under ./tests/data/external/ path there.
```
mimakaev, line 147: This was changed too

mimakaev (approved):
Looks good overall. Some inconsistencies in docstrings.

cooltools/lib/common.py
agalitsyna: In documentation as well, thanks!

cooltools/lib/common.py
agalitsyna: Thanks, docs improved!

.gitignore
```diff
@@ -95,3 +95,6 @@ ENV/
 .ropeproject
 
 tmp/
+
+# Downloaded data
+tests/data/external/
```
nvictus, line 100:
```suggestion
cooltools/datasets
```

agalitsyna: Solved

PR #204, opened 2020-11-17
some changes to cis eigen [will remove soon]

sergpolly:


sergpolly:
to be deleted because of the develop-branch mess, but not yet - wanna make sure all changes have made it here

PR #207, opened 2020-12-01
Logbin refactoring

mimakaev:
Added interpolate_expected and a matching test function.


mimakaev:
@sergpolly  -  do you think this should be merged?

tests/test_expected.py
```diff
@@ -325,14 +343,145 @@ def test_logbin_expected_cli(request, tmpdir):
...
+    exp1, der1, bins = logbin_expected(df)
+    comb, cder = combine_binned_expected(exp1, der1)
+    interp = interpolate_expected(df, comb, by_region=False)
+    max_dev = np.nanmax(
+        np.abs(
+            (interp["balanced.avg"] - df["balanced.sum"] / df["n_valid"])
+            / interp["balanced.avg"]
+        )
+    )
+    assert max_dev < 0.1  # we are now different because we fed combined expected
```
sergpolly, line 408: comment does not belong here I believe - copy/paste mishap ?

tests/test_expected.py
```diff
@@ -325,14 +343,145 @@ def test_logbin_expected_cli(request, tmpdir):
...
+
+    # using chr2 for chr1 works if they are the same
+    interp = interpolate_expected(df, exp1, by_region="chr2")
+    max_dev = np.abs(
+        (interp["balanced.avg"] - df["balanced.sum"] / df["n_valid"])
+        / interp["balanced.avg"]
+    )
+    assert (
+        np.nanmax(max_dev) < 0.1
+    )  # we are now different because we fed combined expected
```
sergpolly, line 430: " we are now different because we fed combined expected" - copy/paste

sergpolly (commented):
sorry for delay @mimakaev ! I've finally tried it and read the code a bit:
 - I'm just asking for comments in the most surprising spots - see comments inside the review
 - not sure if `mat_expected` should return `NaN`-s for the ignored diagonals, for consistency
 - `interpolate_expected` was also a bit "surprising" - maybe we should add a new column with the interpolated value ? rename it ? drop non-interpolated columns ?

cooltools/expected.py
```diff
@@ -920,6 +920,170 @@ def blocksum_asymm(
...
+
+        mask = subdiag >= exclude_diag
+        subhm = subhm[mask]
+        subdiag = subdiag[mask]
+
+        n_valid = np.bincount(subdiag)
+        balanced_sum = np.bincount(subdiag, weights=subhm)
+        diag = np.arange(len(n_valid), dtype=int)
+
+        mask2 = n_valid > 0
```
nvictus, line 1068: Are we dropping diagonals with 0 valid pixels? I thought the expected format should include a record for every relevant diagonal.

mimakaev: Hm. I will think about it. Not 100% sure why I wrote this.

sergpolly: just saw the comment - I've noticed that as well ... I think we should NaN-fill those

sergpolly:
@mimakaev I realized there are more minute issues with `mat_expected` e.g.:
 - it does do double counting in the symmetric case (i'm not sure how to undo it gracefully)
 - if the last diagonal has no valid pixels - it won't be part of the output - i.e. the length of output data-frame wouldn't match between `mat_expected` and `diagsum`
 - when `raw_heatmap` is provided, `mat_expected` filters the same bins in it as it filters in the balanced heatmap - `diagsum` doesn't filter anything
 
these are minor things and double counting does not affect `balanced.sum/n_valid`, but if want to achieve "parity" with the `diagsum` - we'd need to address those
@mimakaev do you mind if I modify `mat_expected` to address those ? also I'd switch from `ar=np.arange(M); diags = np.abs(ar[:,None]-ar[None,:])` to the `LazyToeplitz` ?

sergpolly:
here is a potential version of `mat_expected` https://gist.github.com/64eb38db174c27d7a07cc0b779dee1e6 - cell 6 in the notebook

mimakaev:
Go for it - I won't be using it in the nearest future.

mimakaev:
About double-counting - I'm not sure either. It's an edge usecase anyways...

sergpolly:
Alright ! thank you !
> About double-counting - I'm not sure either. It's an edge usecase anyways...

I meant that it double counts even in the symmetric case - the most used case ...
```python
ar = np.arange(10)
diags = np.abs(ar[:,None] - ar[None,:])
np.bincount(diags.flatten()) # will double count triu and tril diagonals ...
```
this does not affect `balanced.sum/n_valid` because heatmap is symmetric, but would be nice to address
perhaps by just ignoring the `tril` part of the heatmap altogether - not sure yet

tests/test_expected.py
sergpolly: done

cooltools/expected.py
```diff
@@ -920,12 +920,193 @@ def blocksum_asymm(
...
+            else:
+                D_raw = diag_indicator[lo1:hi1, lo2:hi2].flatten()
+                count_sum = np.bincount(D_raw, weights=C, minlength=diag_max)
+
+            count_sum[~mask_per_diag] = np.nan
+            df["count.sum"] = count_sum
+
+        result.append(df)
+
+    return pd.concat(result, ignore_index=True)
```
nvictus, line 1101: I propose to simplify `diagsum_from_array` to work only on one regional submatrix at a time. @sergpolly @gfudenberg what do yout think?

```suggestion
def diagsum_from_array(
    A,
    counts=None,
    *,
    offset=0,
    ignore_diags=2,
    filter_rawmap=False,
    region_name=None
):
    """
    Calculates Open2C-formatted expected for a dense submatrix of a whole
    genome contact map.

    Parameters
    ----------
    A : 2D array
        Normalized submatrix to calculate expected (``balanced.sum``).
    counts : 2D array or None, optional
        Corresponding raw contacts to populate ``count.sum``.
    offset : int or (int, int)
        i- and j- bin offsets of A relative to the parent matrix. If a single
        offset is provided it is applied to both axes.
    ignore_diags : int, optional
        Number of initial diagonals to ignore.
    filter_counts : bool, optional
        Apply the validity mask from balanced matrix to the raw one. Ignored
        when counts is None.
    region_name : str or (str, str), optional
        A custom region name or pair of region names. If provided, region
        columns will be included in the output.

    Notes
    -----
    For regions that cross the main diagonal of the whole-genome contact map,
    the lower triangle "overhang" is ignored.

    """
    if isinstance(offset, (list, tuple)):
        offset1, offset2 = offset
    else:
        offset1, offset2 = offset, offset

    if isinstance(region_name, (list, tuple)):
        region1, region2 = region_name
    elif region_name is not None:
        region1, region2 = region_name, region_name

    A = np.asarray(A, dtype=float)
    if counts is not None:
        counts = np.asarray(counts)
        if counts.shape != A.shape:
            raise ValueError("`counts` must have the same shape as `A`.")

    # Compute validity mask for bins on each axis
    A[~np.isfinite(A)] = 0
    invalid_mask1 = np.sum(A, axis=1) == 0
    invalid_mask2 = np.sum(A, axis=0) == 0

    # Prepare an indicator matrix of "diagonals" (toeplitz) where the lower
    # triangle diagonals are negative.
    lo1, hi1 = offset1, offset1 + A.shape[0]
    lo2, hi2 = offset2, offset2 + A.shape[1]
    ar1 = np.arange(0, hi1 - lo1, dtype=np.int32)
    ar2 = np.arange(0, hi2 - lo2, dtype=np.int32)
    diag_indicator = toeplitz(-ar1, ar2)
    diag_max = max(hi1 - lo1, hi2 - lo2)

    # Apply the validity mask to the indicator matrix.
    # Both invalid and lower triangle pixels will now have negative indicator values.
    diag_indicator[invalid_mask1, :] = -1
    diag_indicator[:, invalid_mask2] = -1

    # Drop invalid and lower triangle pixels and flatten.
    mask_per_pixel = diag_indicator >= 0
    A_flat = A[mask_per_pixel]
    D_flat = diag_indicator[mask_per_pixel]

    # Group by diagonal and aggregate the number of valid pixels and pixel values.
    diagonals = np.arange(diag_max, dtype=int)
    n_valid = np.bincount(D_flat, minlength=diag_max)
    balanced_sum = np.bincount(D_flat, weights=A_flat, minlength=diag_max)

    # Mask to ignore initial diagonals.
    mask_per_diag = diagonals >= ignore_diags

    # Populate the output dataframe.
    # Include region columns if region names are provided.
    # Include raw pixel counts for each diag if counts is provided.
    balanced_sum[~mask_per_diag] = np.nan
    df = pd.DataFrame({
        "diag": diagonals,
        "n_valid": n_valid,
        "balanced.sum": balanced_sum,
    })

    if region_name is not None:
        df.insert(0, "region1", region1)
        df.insert(1, "region2", region2)

    if counts is not None:
        C_flat = counts.flatten()

        # Either count everything or apply the same filtering as A.
        if filter_rawmap:
            C_flat = C_flat[mask_per_pixel]
            count_sum = np.bincount(D_flat, weights=C_flat, minlength=diag_max)
        else:
            D_raw = diag_indicator[lo1:hi1, lo2:hi2].flatten()
            count_sum = np.bincount(D_raw, weights=C_flat, minlength=diag_max)

        count_sum[~mask_per_diag] = np.nan
        df["count.sum"] = count_sum

    return df
```

nvictus: > is the rationale - to remove slightly mysterious _preprocess_regions

The rational is to make it a simpler function with a simpler goal. The caller can loop over regions and name them whatever they want.

> offset changed its meaning  Am I missing something about non-square A as input ?

Yes, the idea is that you pass in the submatrix as input (generally that is something that can fit in memory). `offset` refers to the positional offset of the input submatrix relative to the whole-genome parent matrix. i.e. what you get from `clr.offset(region)`.

> Also rename filter_rawmap to filter_counts as in the docstring

Thank you.

sergpolly: Oh, I see - so it's really meant to "consume" arbitrary rectangles ... including e.g. inter-arm

But I'm still struggling with the `toeplitz(-ar1, ar2)` - I haven't tried to run it yet - but it looks like that the way you define `ar1` and `ar2` (starting from `0`) - would "force" the "tril" elements of `A` to be ignored - even with some `offset` , no ?

for this to work properly `ar1` shouldn't be always negative - it should negative only for the tril "overhang" ...

maybe sorta kinda like that:
```python
    lo1, hi1 = offset1, offset1 - A.shape[0] # decreasing - not sure if syntax is sound
    lo2, hi2 = offset2, offset2 + A.shape[1]
    ar1 = np.arange(lo1, hi1, step= -1, dtype=np.int32) # hi1 < lo1 now
    ar2 = np.arange(lo2, hi2, dtype=np.int32)
    diag_indicator = toeplitz(ar1, ar2) # negativity of ar1 decided upstream
```

nvictus: Yes, you're right. My mistake. I think it should be like this:

```python
    lo1, hi1 = offset1, offset1 + A.shape[0]
    lo2, hi2 = offset2, offset2 + A.shape[1]
    ar1 = np.arange(lo1, hi1, dtype=np.int32)
    ar2 = np.arange(lo2, hi2, dtype=np.int32)
    diag_indicator = ar2[:, np.newaxis] - ar1[np.newaxis, :]  # or whatever the toeplitz equivalent is
```

We also need to get the lower and upper bounds of diagonals to calculate summaries for. These diagonals refer to the whole genome parent matrix, not the submatrix.

```python
  diag_min = max(lo2 - hi1, 0)
  diag_max = hi2 - lo1
```

And modify the bincount invocations accordingly:

```python
diagonals = np.arange(diag_min, diag_max, dtype=int)
n_valid = np.bincount(D_flat, minlength=diag_max - diag_min)
```

It needs testing, regardless.

Issue #168, opened 2020-06-26
Error in ObsExpSnipper?


Mittmich:
We started migrating our [labinternal wrappers](https://github.com/gerlichlab/ngs) for cooltools to a new version of cooltools (0.2.0 to 0.3.2) and noticed that our pileups are not working anymore. I get a `ValueError` in `ObsExpSnipper` when the individual groups are accessed:https://github.com/mirnylab/cooltools/blob/d9f43d3120d72e242c00f029c655887cc9f30681/cooltools/snipping.py#L359

A notebook to reproduce the error with our testdata is available [here](https://github.com/gerlichlab/ngs/blob/new_cooltools/Cooltools_error.ipynb). The same code with an older version that worked can be found [here](https://github.com/gerlichlab/ngs/blob/new_cooltools/Cooltools_old_error.ipynb).

The commit that changed the behavior is this one: https://github.com/mirnylab/cooltools/commit/86f2636c25f0351d7056d3f19bbe629d6c913017#diff-2c39b1e7b7e8511cda2d9ee8ddcc0335

Is this expected behavior and I need to change the way I feed the data, or is this a bug?

mimakaev:
I remember seeing this code, and I even remember fixing this bug, but I probably forgot to commit it because I can't find my fix. It was something easy. So this is definitely a bug. I'll see if I can recall the solution.

mimakaev:
Submitted a PR that I believe corrects this issue. Not sure how and why this happened in the first place... but I should wait from users of this code to report perhaps. Maybe someone is using regions as a tuple as a first column of expected dataframe? https://github.com/mirnylab/cooltools/pull/171

Mittmich:
Thanks a lot! I was also thinking that this might be a possibility, but the `region` and `region2` arguments that are passed to `.snipping` come from `bioframe.parse_region_string`, which is always a tuple of `(chrom, start, end)`, so the above code will always select `chrom` because `len(self.regions_columns)` is always `> 0`. This might work if `self.regions_columns` are a single chromosome name, but I think it will fail in all other cases.

valtona:
Hello,
When using snipping for dot pileups I am now having this error:

```
KeyError                                  Traceback (most recent call last)
<ipython-input-33-b559c633342e> in <module>
      4     expected = pd.read_table(f'/data/.../expected/{long_names[cond]}.{binsize//1000}kb.expected.cis.tsv')
[...]

Issue #177, opened 2020-08-07
KeyError in ObsExpSnipper for input expected tsv


zhaoweiyu-github:
Hi, although you fixed bugs reported for calculated expected dataframe (have chrom,start,end columns) [Error in ObsExpSnipper](https://github.com/mirnylab/cooltools/issues/168) but the new version would report KeyError for input expected tsv generated by cooltools compute-expected (only have chrom column but not start,end column).

Phlya:
Yeah, it's the same in ExpectedSnipper. Example:

https://gist.github.com/Phlya/0cd17b6d4f9a1245bef8d5991fce7b77

Phlya:
So I'm not sure what format of region would even work at all for expected files with just "chrom" (old style) or just "region" (future style) column.

Phlya:
This should all be finally fixed in the current pre-release (which will be released imminently), yay!

PR #223, opened 2021-03-12
Fix expected in snipping

Phlya:


Phlya:
Should fix https://github.com/open2c/cooltools/issues/168 and https://github.com/open2c/cooltools/issues/177

PR #234, opened 2021-04-12
Correct two cases where format strings were missing

folded:


Phlya:
Nice little catch, thank you!

Issue #219, opened 2021-03-02
Slow expected calculation


Phlya:
@sergpolly made functions for calculating expected flexible back in summer, which slowed them down a lot. This needs to be fixed to make them great again!

sergpolly:
Anton's proposal:
1. if we want to keep supports overlapping (ultimate flexibility, but exotic case) we can improve complexity of the algorithm by first iterating over pixels and assigning them to different supports . Right now we are going over pairs of regions, assigning supports, pre-calculating sums and then repeating it for a different pair of regions over and over again. `~P*R` complexity  (`P`-# of pixels, `R` - number of 2D regions to handle) ? isn't it `P*R` either way ?
2. if we can guarantee that the regions aren't overlapping - or, better, - there aren't any "support region" clashes - every pixel belong to a different "support" region - than this could be optimized further ... the before-summer performance could be restored

PS
 - I'm afraid scenario 2 is only possible using the `assign_support_2D` that @mimakaev was proposing that same summer
 - scenario 1 is clearly an overkill of "flexibility" - expected for regions overlapping in 2D (sharing pixels) - is exotic, agreed
 - problem was that the "summer" version of `diagsum_asymm` was only capable of processing 2D regions that did not share even a single "side" - i.e. it was "illegal" to put regions like `(chr1p, chr1p), (chr1p, chr1q), ...` as input
 - also, need to explore this branch of `assign_support` further ... https://github.com/open2c/cooltools/blob/53ce4fc11d046e84ca36f3591e9e5d548de9f7b2/cooltools/lib/common.py#L44

Phlya:
How about implementing `assign_supports` using bioframe? Seems like it would be faster without a loop over all supports. And should be relatively easy to turn into `assign_supports_2D` with correct treatment of overlapping regions...

Phlya:
Just wanted to say, we basically implemented the same functionality in snipping using bioframe, maybe it'll help: https://github.com/open2c/cooltools/blob/4c41f419c2e303b01a42238009df2fbfc60303dc/cooltools/snipping.py#L77

sergpolly:
Yes - loop over all `supports/regions/views`is a disaster - it allows regions to be whatever (overlapping, duplicated etc etc) - but it leads to massive slowdowns...

Without some flexible and fast assign_supports/regions_2D that can assign regions non-overlapping in 2D, we'd have to revert back to special cases.

[...]

PR #239, opened 2021-04-16
Faster expected calculation [WIP]

Phlya:
First try to fix expected to work FAST (fix for https://github.com/open2c/cooltools/issues/219). Just fixed the `diagsum` function for now.

What do we think about this kind of solution? It works very fast, but I haven't compared the actual output yet.


Phlya:
Following some further comparisons, it actually doesn't seem like this version is any faster.
However using a smaller chunksize seemed to make expected work much better in the context of limited memory.

PR #245, opened 2021-05-10
Add regions to compartments CLI

Phlya:
Simply added a `regions` argument to `call-compartments` and pass it into the API when asking for cis contacts (regions for trans are not implemented in the API)


sergpolly (approved):
- Modified reading regions in call-compartments to be similar to other CLI tools
 - call-dots has --out-prefix now, instead of output-calls - output files are explained better, and the option is required
 - some black here and there
 - test_call_dots checks is the output files exist

PR #252, opened 2021-06-30
Fix assign regions bug (fixes #251)

### LINKED ISSUE #251: deprecate cooltools copy of assign_regions, switch to bioframe version
In cases where `assign_regions` is used the assign supports to features that span multiple supports, supports are assigned incorrectly. I created a notebook that exemplifies this behavior [here](https://colab.research.google.com/drive/11IRO_Pb9zkoIFac5-ueE9CAjDYRil76r?usp=sharing). I had a look at the code of `assign_regions` and I  think the problem is when [duplicate rows are dropped](https://github.com/open2c/cooltools/blob/9094c0498dd7a4c790be6909f3d051cb0201112b/cooltools/lib/common.py#L34), the column names of the entire overlap region are used, where the multiply assigned supports will differ, so duplicates are not dropped.

Mittmich:
**Fixes bug mentioned in [this issue](https://github.com/open2c/cooltools/issues/251).**

The problem is that when a feature spans multiple supports, the `cooltools.snipping.assign_regions` function assigns supports incorrectly (see [this notebook](https://colab.research.google.com/drive/11IRO_Pb9zkoIFac5-ueE9CAjDYRil76r?usp=sharing) for an example). The error arises in the step where columns are selected for [defining duplicates](https://github.com/open2c/cooltools/blob/9094c0498dd7a4c790be6909f3d051cb0201112b/cooltools/lib/common.py#L34). Here, all columns of the overlap dataframe are taken, but since this dataframe also contains the assigned supports, no duplicates will be removed. I created a test case in `test_snipping.py` that recapitulates this problem.

I suggest a fix by explicitly defining the columns that should be used to search for duplicates (`["chrom_1", "start_1", "end_1"]` in the on diagonal case and `[f"chrom{idx}_1", f"start{idx}_1", f"end{idx}_1"]` in the off-diagonal case).


Phlya:
Thank you for bringing this up and proposing a solution. We were planning to migrate that code to this new bioframe function: https://github.com/open2c/bioframe/blob/d20735f11b40e987e8f682d95998603622b614fc/bioframe/ops.py#L1755

Do you think you could test if this method also has this issue?

Mittmich:
[Just tested](https://colab.research.google.com/drive/1M4e2qmbm_wiE7klOpjCysDtDE9nLFukG?usp=sharing) the `assign_view` function from bioframe and it does not have this problem. Thanks!

Phlya:
Excellent! If you would like to change your PR to introduce `assign_view` to cooltools, that would be perfect :)

Mittmich:
Sure, I would like to do that! I assume want to keep the original interface to `cooltools.snipping.assign_regions`? So I would just use that as a wrapper and replace the logic with `assign_view`, or where you thinking of a different implementation?

Phlya:
I think that sounds good!

Mittmich:
So I looked into this more closely and realized that `bioframe.assign_view` is only available in `bioframe` version 0.2.1. The problem is that the current version of `cooltools` makes use of `bioframe.parse_regions`, which is not availalbe in `bioframe` verions 0.2.1. As far as I can tell, `bioframe.construction.make_viewframe` covers that functionality and I could replace it. However, I was wondering what your plans are with regards to updating `cooltools` to use `bioframe` 0.2.1?

gfudenberg:
Hey Michael-- my hope is that we will freeze bioframe functionality & bump the version to 0.3.0 within the next week or so (you can see the almost-complete checklist on the 0.3.0 roadmap issue in bioframe). Then we can sync cooltools to rely on the updated bioframe!

gfudenberg:
closing this PR, since assign_view is now in bioframe and the cooltools assign_region is deprecated

Issue #254, opened 2021-08-05
streamline cooltools API - assume "regions" are "viewframe", retire parsing


sergpolly:
proposal to streamline/simplify internal cooltools API: any `regions` argument has to be an instance of `bioframe.viewframe` , so we can retire any regions-parsing that was done internally and replace it with something like that:
```python
try:
    bioframe.is_viewframe(regions, raise_error=True)
except (TypeValue, ValueError) as error:
    print(error)
    raise "regions has to be an instance of bioframe.viewframe"

# or just this:
bioframe.is_viewframe(regions, raise_error=True)

# or this:
if not bioframe.is_viewframe(regions, raise_error=False):
    raise "regions has to be an instance of bioframe.viewframe"

# not sure which is best
```

sergpolly:
from the aug 17 meeting - regarding non-CLI:
 - try to allow some flexibility for the "user"-facing API functions (e..g allow regions=None, or any other format supported by make_viewframe)
 - for more internal/non-public/lower level functions we can stick with requiring (or even make no checks !) strict `is_viewframe` check.

sergpolly:
from aug 17 meeting - regarding CLI:
use following pattern to read/parse through a user provided path to `regions/view`:
```python
if regions is None:
    # use entire chroms
else:
    try:
        regions_df = bioframe.read_table(regions_buf,schema="bed4")
    except  whatever:
        regions_df = bioframe.read_table(regions_buf,schema="bed3")
try:
    regions_df = bioframe.make_viewframe(bioframe.read_table("xxx",schema="bed3"))
except ValueError as e:
    raise RuntimeError("ypu've provided shitty regions file") from e
```

agalitsyna:
Hi, small update in the [recent commit](https://github.com/open2c/cooltools/commit/ef4a1c6f2ed2daf6481d8e7d3b0cb1509c5e7cbb), what's been done: 

- CLIs supporting viewframes: call-compartments, call-dots, compute-expected, compute_saddle
- APIs supporting viewframes: eigdecomp, expected, saddle, snipping
- Corresponding tests updated.

Noticed problems: 

- [ ] asym expected and trans eigdecomp require repeated entries in regions, which is not supported by bioframe.make_viewframe. This actually requires discussion with @sergpolly @gfudenberg 
This currently fails the tests of asym expected and trans compartments/saddles.

[...]

PR #260, opened 2021-08-24
Rename regions to view everywhere

Phlya:
I left the old names in places where we need to fix the code anyway, e.g. asym expected

### DIFF
--- a/cooltools/expected.py
+++ b/cooltools/expected.py
@@ -541,7 +541,7 @@ def _diagsum_symm(clr, fields, transforms, regions, span):
 
 def diagsum(
     clr,
-    regions,
+    view_df,
     transforms={},
     weight_name="weight",
     bad_bins=None,
@@ -557,8 +557,8 @@ def diagsum(
     ----------
     clr : cooler.Cooler
         Cooler object
-    regions : viewframe or sequence of genomic range tuples
-        Support regions for intra-chromosomal diagonal summation
+    view_df : viewframe or sequence of genomic range tuples
+        Support view_df for intra-chromosomal diagonal summation
     transforms : dict of str -> callable, optional
         Transformations to apply to pixels. The result will be assigned to
         a temporary column with the name given by the key. Callables take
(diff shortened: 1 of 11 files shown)


cooltools/expected.py
```diff
@@ -557,8 +557,8 @@ def diagsum(
     ----------
     clr : cooler.Cooler
         Cooler object
-    regions : viewframe or sequence of genomic range tuples
-        Support regions for intra-chromosomal diagonal summation
+    view_df : viewframe or sequence of genomic range tuples
```
agalitsyna, line 560: Looks like diagsum does not support tuples anymore

cooltools/saddle.py
```diff
@@ -73,7 +73,7 @@ def digitize_track(binedges, track, regions=None):
         edges. See encoding details in Notes.
     track : tuple of (DataFrame, str)
         bedGraph-like dataframe along with the name of the value column.
-    regions: viewframe, sequence of str or tuples
+    view_df: viewframe, sequence of str or tuples
```
agalitsyna, line 76: No backward compatibility: looks like view_df cannot be a sequence of str or tuples anymore

cooltools/saddle.py
```diff
@@ -283,7 +286,7 @@ def make_saddle(
     contact_type : str
         If 'cis' then only cis interactions are used to build the matrix.
         If 'trans', only trans interactions are used.
-    regions : viewframe, sequence of str or tuple, optional
+    view_df : viewframe, sequence of str or tuple, optional
```
agalitsyna, line 289: Looks like view_df cannot be a sequence of str or tuple anymore?

agalitsyna (changes requested):
I just noticed (unrelated to the work that Ilya has done) that at least three functions have lost backward compatibility with old code. For example, if someone uses old code with diagsum, it won't support sequences of tuples as input. 
I would retain backward compatibility of diagsum/digitize_track/etc. If we don't want it, the docstrings are to be changed.

Phlya:
I thought that was the decision, that we require a proper viewframe from the user now? https://github.com/open2c/cooltools/issues/254

sergpolly:
@Phlya that what I was wondering, but @nvictus suggested that it should be phased out (in) gradually at least ...

I wonder if we could replace those `assert` with something like this:
```python
try:
     bioframe.is_viewframe(view_df, ...,raise_errors=True)
     bioframe.is_contained(...,raise_errors=True)
except WhatEverError as e:
    # https://docs.python.org/3/library/warnings.html#warnings.warn
    warning.warn("view_df has to be a proper viewframe from next release on ...", DeprecationWarning, stacklevel=2)
    # try to make viewframe from whatever was provided
    view_df = bioframe.make_viewframe(view_df)
```

Issue #262, opened 2021-08-24
v0.5 roadmap


Phlya:
#### API and CLI:
- [x] https://github.com/open2c/cooltools/issues/291 @agalitsyna 
- [x] synchronize CLI with new user-facing API functions
  - [x] consistent naming of user-facing function: no prefixes, bury modules under .api.
  - [x] consistent CLI naming: no prefixes
- [x] split expected CLI into cis and trans
- [x] discuss logbinning -> log_smoothing
- [x] split eigs-cis and eigs-trans in the CLI, its more than just a flag
- [x] include smoothing into user-facing expected-cis, update notebooks to use this

#### Code updates:
- [x] ~~several places in the code pass tuples to make_viewframe instead of lists, which is not recognized by `bioframe.from_any`.~~ `bioframe.from_any()` now supports bedframe construction from tuples
- [x] Rename regions to views, if possible (see #254)
- [x] asym expected require repeated entries in regions, which is not supported by `bioframe.make_viewframe`. This currently fails the tests of asym expected and `test_trans_saddle_cli`. see #254. 
   Options:
   - Single 1D viewframe, rewrite functions to work on all pairwise combinations of regions
   - Allow repeated entries in viewframes with an argument in `is_viewframe`
   - Migrate to 2D viewframes (requires new functions and definitions in bioframe)
- [x] In call-dots and saddle we take UCSC-named regions and expected as a requirement. Rewrite to avoid relying on that 
  - [x]  https://github.com/open2c/cooltools/blob/viewframe_update/cooltools/cli/compute_saddle.py#L325 
  - [x]   https://github.com/open2c/cooltools/blob/viewframe_update/cooltools/cli/call_dots.py#L31
- [x] Fix coverage by removing division by 2 (#255) if the cooler is upper triangular
- [x] update columns for expected to correspond to standard #217
- [x] rename to clr_weight_name #272
- [x] make saddle CLI save sums and counts separately for flexible "strength" calculation #261
- [x] restore compute-expected performance for cis-symmetric and trans-pairwise: refactor _diagsum_symm and re-introduce blocksum_pairwise #219
- [x] Distinguish user-facing / public functions vs "internal" functions, e.g. by explicit import of the former into the top namespace #258

#### Checks
- [x] checks.is_expected to validate an expected table has correct columns, dtypes, etc. https://github.com/open2c/cooltools/issues/217

#### ReFactoring
[...]

Issue #257, opened 2021-08-17
migrate "regions" to "views" in the API and command line interfaces


sergpolly:

Phlya:
Mostly addressed in #260 - except a couple of places where the concept of 1D views is limiting us at the moment.

Issue #255, opened 2021-08-12
divide by 2 issue in cooltools.coverage


gfudenberg:
if pixels are upper-triangular, then dividing by 2 should not be required, i.e. we should remove:
> [covs = covs / 2](
https://github.com/open2c/cooltools/blob/7bb960848f23dc187078d5c84287d3d073adb999/cooltools/coverage.py#L47
)

![image](https://user-images.githubusercontent.com/19657573/129250840-46227ca5-dac0-42a9-b4c4-8eb068ac327d.png)

sergpolly:
that bothered me some time ago, but I was convinced this was correct: https://github.com/open2c/cooltools/issues/99#issuecomment-543974052

I'll try to understand it deeper again

sergpolly:
so yeah , i think divide by 2 is correct way, because thouse `bincount`-s are indeed double counting every pixel
![Screenshot from 2021-09-01 19-42-10](https://user-images.githubusercontent.com/6790270/131759439-01793ab6-1b70-4acb-9a31-a9392bec9036.png)

e.g. that pixel `[2,3]` in the figure - it'll "participate" in the summing up counts for row `2` AND row `3` - so you'll count it twice ...
So , to keep the `sum(coverage)` same as the total number of pairs/ligations that went into creating this matrix - every pixel should contribute `0.5*count` to every row instead of `count`

PS `sum(coverage)` wouldn't match total number of "pairs" because we ignore diagonals, etc

sergpolly:
that bothered me so much, i made some slides about it - https://github.com/open2c/cooltools/issues/99#issuecomment-569380032

check slides 20, 21, ... for the demonstration of why every pixel contributes 2x to the coverage at the end

agalitsyna:
Hi, Sergey @sergpolly, I think I understand your point. You mean that if we  sum up all the coverages without dividing by 2, we will count the upper triangular matrix twice, and it won't equal the total number of contacts in the matrix. This is true, and the slides 20-21 are actually an excellent illustration of the point! Never thought of such a beautiful Hi-C binning representation before. 

Then my question is, when in Hi-C analysis do actually need to sum up all the coverages? If we want total number of contacts, we take them directly from the sparse matrix. If I want the bin coverage, we usually assume *the total number of pairs in which each region participates*. In the absence of the main diagonal, the answer is straightforward:
`covs[1] += np.bincount(pixels["bin1_id"], weights=pixel_weights, minlength=n_bins)
[...]

PR #265, opened 2021-08-31
Viewframe update

agalitsyna:
Cooltools transfer to bioframe.viewframe
Key tasks/issues resolved: 
- 1-5 from https://github.com/open2c/cooltools/issues/262
- https://github.com/open2c/cooltools/issues/255
- https://github.com/open2c/cooltools/issues/254
- https://github.com/open2c/cooltools/issues/257
- https://github.com/open2c/cooltools/pull/260


cooltools/dotfinder.py
```diff
@@ -1341,7 +1342,9 @@ def scoring_step(
             # "create_cooler" API needs more consistency:
             # https://github.com/mirnylab/cooler/issues/149
             #  manually describing "data"-columns (except for bin_ids).
-            columns = ["count",] + ["la_exp." + k + ".value" for k in kernels]
+            columns = [
+                "count",
```
sergpolly, line 1346: sometime `black` does strange things - like this one ...
I hope, at least , that it stays consistent from run to run, from version to version

tests/test_compartments_saddle.py
```diff
@@ -120,12 +117,8 @@ def test_trans_saddle_cli(request, tmpdir):
...
     runner = CliRunner()
     result = runner.invoke(
-        cli, [
-            'call-compartments',
-            '--contact-type', "trans",
-            '-o', out_eig_prefix,
-            in_cool
-        ]
+        cli,
+        ["call-compartments", "--contact-type", "trans", "-o", out_eig_prefix, in_cool],
```
sergpolly, line 121: `black` ...

tests/test_coverage.py
```diff
@@ -0,0 +1,26 @@
...
+import numpy as np
+
+
+def test_coverage_symmetric_upper(request):
+    # perform test:
+    clr = cooler.Cooler(op.join(request.fspath.dirname, "data/CN.mm9.1000kb.cool"))
+    cov = cooltools.coverage.get_coverage(clr, ignore_diags=2, chunksize=int(1e7))
+
+    # Test that minimal coverage is larger than 0.5
+    assert cov[cov > 0].min() >= 1
```
sergpolly, line 14: the debate is still pending for this

sergpolly (changes requested):
This looks great ! Should be ready to merge after we fix maybe a couple of tiny items:
1. `RuntimeError`  -> `ValueError` in a couple of places
2. remove some comments related to the issues with `regions, regions2` not being unique for trans and some asymmetric cases
3. resolve the debate regarding coverage - i.e. do we do `cov` or `cov/2` ?

We cannot do much with `black`, can we ? I feel like we ran `black` on the entire codebase before, but some weird formatting changes are still creeping in. I don't understand it fully. Maybe we could agree on doing `black` as separate PRs every now and then, to make the changes looks smaller ? Also , any chance we could have slightly different `black` parameters - so that we keep re-`blacking` over and over ?

cooltools/coverage.py
```diff
@@ -44,8 +44,6 @@ def _get_chunk_coverage(chunk, pixel_weight_key="count"):
     covs[1] += np.bincount(pixels["bin1_id"], weights=pixel_weights, minlength=n_bins)
     covs[1] += np.bincount(pixels["bin2_id"], weights=pixel_weights, minlength=n_bins)
 
-    covs = covs / 2
```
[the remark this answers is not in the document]
gfudenberg: discussion 9/7/21:
-- lets go with removing /2 everywhere, including the [special case]( https://github.com/open2c/cooltools/blob/ff066aca1cd1569c119cfe74efdb54beee580e20/cooltools/coverage.py#L111), and update the docstring to clarify that the sum( total_coverage) from this function is 2x the number of reads contributing to the cooler. The storage mode is apparently for a different type of coolers (e.g. RNA vs DNA) that are not supported currently by cooltools.
    
Calculate the sums of cis and genome-wide contacts (aka coverage aka marginals) for  a sparse Hi-C contact map in Cooler HDF5 format. Note that the sum( `tot_raw_cov`) from this function is two times the number of reads contributing to the cooler as each side contributes to the coverage.

tests/test_coverage.py
sergpolly: Based on the discussion 9/7/2021 -> this is correct

sergpolly (approved):

Issue #258, opened 2021-08-17
discriminate between user-facing and non-public functions


sergpolly:
user-facing functions can be imported into the base namespace, e.g.:
```python
from cooltools import *
```
user-facing functions should have the following format:
```python
get_XYZ(
  clr, 
  tracks1=Y1, track2=Y2, .... (input data)
  cutoffs1=Z, cutoffs2=Z2, .... (input parameters)
  nproc=?, chunksize=?, ... (execution specification)
)
```

list of potential user-facing functions by module:

`cooltools.coverage`
- `get_coverage`

`cooltools.directionality` and `cooltools.insulation`. consider refactoring into one module, with different scores. Implement as flexible convolution along the diagonal. @golobor 
- `get_directionality`
- `get_insulation`
- `find_boundaries`

`cooltols.dotfinder` 
- `get_dots` (need to merge pieces together as done in CLI)

`cooltools.eigdecomp` @nvictus @e-navarrete
- `cooler_cis_eig` --> `get_cis_eig`, internal cis_eig should get an underscore
- `cooler_trans_eig` --> `get_trans_eig`, internal should get underscore etc.

`cooler.expected`. refactor diagsum, diagsum_asym (cis-arms), blocksum_asymm (trans)... consider refactoring the CLI for expected into two separate functions. @sergpolly 
- `get_cis_expected` (supports views (i.e. arms), flags for smoothing=True by default, aggregate=True by default) 
- `get_trans_expected`
- `contacts_vs_distance` (also includes derivative) @golobor 

`cooler.saddle`: needs refactoring to make the following work @gfudenberg  
- ~~`make_saddle` --> `get_saddle`. needs to be changed (or wrapped) to accept a cooler as input.~~
- `saddle_strength`. update docstring to clarify it relies on make_saddle output.

`cooler.sample`:
- `sample_cooler`

`cooler.snipping`: @Phlya 
- ????

This would be a way to distinguish the "public" functions vs more internal/lowlevel ones. This has the bonus where the lower-level functions can be freed from tedious parsing and sanitation of input arguments.

gfudenberg:
are `compute_scaling`, `contact_areas`, and `_contact_areas` the functions for reads that should be moved out of expected? cc @golobor @nvictus

gfudenberg:
done, right? cc @golobor

golobor:
seems so! expect for saddle. saddleplot

gfudenberg:
ok, closing this & making an informative issue for saddleplot

Issue #42, opened 2018-09-25
binedges are not used by make_saddle; use max(digitized) instead


golobor:
https://github.com/mirnylab/cooltools/blob/34ba9a347010b614f483871baa74d29dbe23dd9d/cooltools/saddle.py#L216

gfudenberg:
I think they are passed b/c the "right outlier values" bin could be empty and hence max(digitized) would give a number of bins that is off by one... would it be strange to make digitized a categorical so that information is passed in the dataframe?

related to this, though, digitized currently puts the max(track) in the "outlier bin", 
![image](https://user-images.githubusercontent.com/19657573/132107319-f135c4fb-aec4-4e73-a53c-6685fc9116a3.png)

https://github.com/open2c/cooltools/blob/ff066aca1cd1569c119cfe74efdb54beee580e20/cooltools/saddle.py#L66

since np.digitize ``bins[i-1] <= x < bins[i]`` for all bins. 

should we modify this to be <= bins[-1]?

golobor:
I agree, max(digitized) should belong to the last bin!

gfudenberg:
great-- this is addressed now, closing!


# re-assign values equal to the max value to bin n
digitized.loc[
        digitized[track_value_col] == np.max(binedges), track_value_col
    ] = n_bins

PR #273, opened 2021-09-10
Saddle refactor

gfudenberg:
updates & refactoring & tests for `cooltools.saddle` with https://github.com/open2c/cooltools/issues/258 in mind. 

- `get_saddle()` replaces `make_saddle`, and directly accepts (clr, expected, digitized_track), which will clarify the API.
- obs/expected fetchers are now underscored and understood as helper functions for `get_saddle`.
- `get_digitized` returns a dataframe and binedges instead of a histogram, and merges in what had previously been used in the CLI for bin creation. 
- digitized track is now a **categorical**

 in the end, I decided that two separate functions for cis vs trans wasn't 100% necessary, and that it wouldn't be overly cumbersome to have the digitized track as a requirement for get_saddle, since it now contains the bin information in its categories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Our contributing guide can be found [here](https://github.com/open2c/cooltools/b
 
 The following are required before installing cooltools:
 
-* Python 3.4+
+* Python 3.7+
 * `numpy`
 * `cython`
 
--- a/cooltools/cli/compute_saddle.py
+++ b/cooltools/cli/compute_saddle.py
@@ -93,7 +93,7 @@ from . import cli
     show_default=True,
 )
 @click.option(
-    "--weight-name",
+    "--clr-weight-name",
     help="Use balancing weight with this name.",
     type=str,
     default="weight",
@@ -169,7 +169,7 @@ def compute_saddle(
     quantiles,
     range_,
     qrange,
-    weight_name,
+    clr_weight_name,
     strength,
     view,
     out_prefix,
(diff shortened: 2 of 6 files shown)


gfudenberg:
forgot to mention, I also go with @itsameerkat 's suggestion of n_bins+1 in `get_digitized`, so that if the bins for low outliers and high outliers are removed (i.e. bins #0 and #n+1) then there are n_bins left. So I think we can probably close
https://github.com/open2c/cooltools/pull/152/files

I didn't change the trim_outliers to True by default, but it could be a good idea.

If we do, we should also decide what to do with the max value of track, which currently gets cast into the high outliers by np.digitize-- see my comment: https://github.com/open2c/cooltools/issues/42

gfudenberg:
following discussion on 9/15 the following has been updated:
+ update arguments to 'clr_wieght_name', 'expected_value_col' in `cli.compute_saddle get_saddle()` to take click inputs
+ add input checking for `get_digitized()`, `_make_bin_edges` (e.g.` _is_track`)
+ change range_ to vrange for value range, including in cli click options (now --vrange)
+ remove quantiles form _make_binedges, make qrange & vrange mutually exclusive, raise error if lo>hi.
+ add a deprecation warning to saddleplot. 
+ add saddlecounts to_save for cli.
+ added filter to trans saddles list of supports to remove regions with idential chrom in view_df
+ clean up the bins[-1] by re-assigning max values (np.finfo(dtype).eps was unstable)

PR #274, opened 2021-09-17
Speedup expected

sergpolly:
- reintroduced efficient diagsum_symm and blocksum_pairwise (with new bioframe checks it's so easy validate regions!)
 - cleaned up compute_expected CLI and updated in accordance with new API
 - modified tests to have bin-aligned regions - it is not fair otherwise

Observations 
 - old `assign_supports` is much faster than anything `bioframe.overlap`-based, as it is just logical comparisons on 6 columns in a `pixels` tables . Can be further speed up operating directly on `bin1_id, bin2_id` and `np.searchsorted`
 
Questions:
 - do we need an efficient `diagsum_pairwise` or `diagsum_interleaved` e.g. for inter-arm , and other inter-region intra-chromosomal expected ?
 - should `blocksum_pairwise` return only inter-chromosomal results or *all* of the results (e.g. inter-arm average if arms were provided as `view_df`)?

### DIFF
--- a/cooltools/expected.py
+++ b/cooltools/expected.py
@@ -542,10 +542,12 @@ def make_block_table(clr, regions1, regions2, weight_name="weight", bad_bins=Non
 
 def _diagsum_symm(clr, fields, transforms, regions, span):
     """
-    calculates diagonal summary for a collection of
-    square symmteric regions defined by regions.
-    returns a dictionary of DataFrames with diagonal
-    sums as values, and 0-based indexes of square
+    calculates diagonal/distance summary for a collection of
+    square symmetric blocks defined by the "regions".
+
+    Return:
+    dictionary of DataFrames with diagonal/distance
+    sums for the "fields", and 0-based indexes of square
     genomic regions as keys.
     """
     lo, hi = span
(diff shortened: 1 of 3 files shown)


cooltools/expected.py
```diff
@@ -987,6 +1377,287 @@ def blocksum_asymm(
     )
 
 
+# user-friendly wrapper for diagsum_symm and diagsum_pairwise - part of new "public" API
+def get_cis_expected(
+    clr,
+    view_df=None,
+    symmetric=True,
```
sergpolly, line 1384: make it return either symmetric expected only, or return "everything" cis - e.g. intra- and inter-arm expected.

cooltools/expected.py
```diff
@@ -987,6 +1377,287 @@ def blocksum_asymm(
...
+    clr : cooler.Cooler
+        Cooler object
+    view_df : viewframe
+        a collection of genomic intervals where expected is calculated
+        otherwise expected is calculated for full chromosomes.
+    symmetric: bool
+        When True, view_df defines symmetric on-diagonal regions,
+        otherwise expected is calculated for every intra-chromosomal
+        combination of regions in view_df
+    balance : bool or str
```
sergpolly, line 1412: turn it into `clr_weight_name` that is `weight` by default. And `clr_weight_name=None` would do the unbalanced ...

cooltools/expected.py
sergpolly: intra_only =True- people like it
intra inter
matching_pairs
...

cooltools/expected.py
```diff
@@ -987,6 +1377,287 @@ def blocksum_asymm(
...
+                view_df,
+                transforms=transforms,
+                weight_name=clr_weight_name,
+                bad_bins=None,
+                chunksize=chunksize,
+                ignore_diags=ignore_diags,
+                map=map_,
+            )
+            # conform with the new expected format
+            # where each region is 2D, even if symmetric:
```
sergpolly, line 1500: actually hide the format transformation inside of `diagsum`. And consider making `region1/2` a categorical for lightness

PR #298, opened 2021-11-05
Pileups api

Phlya:


sergpolly:
what are these `.ini.bak` files ? looks like some IDE settings ... do we need them ?

cooltools/cli/compute_pileup.py
```diff
@@ -0,0 +1,260 @@
...
+
+    features_df = pd.read_table(
+        buf, comment="#", usecols=default_cols, dtype=dtypes, verbose=verbose, **kwargs
+    )
+
+    ###### Define view for cis compartment-calling
+    # use input "view" BED file or all chromosomes mentioned in "track":
+    if view is None:
+        # Generate viewframe from clr.chromsizes:
+        view_df = bioframe.make_viewframe(
```
sergpolly, line 181: you can do `make_cooler_view(clr)` for that

sergpolly: or maybe we could keep it as is - and I'll modify it using new functions from .lib.common afterwards

Phlya: I can update, thanks!

sergpolly:
another problem is that - right now you have all of these commits from my branch which are not final, and they are UN-doing some of the changes that were merged in the master
I'm not sure how to resolve that...

Phlya:
But I merged master into this branch today before commiting everything...  AndI resolved all conflicts in favour of master

Phlya:
I don't think there are any changes in the final diff that undo anything in master.

Phlya:
```- tests run OK with min_diag==None```

Wow I'm surprised! Seems like any number is considered less than None, so then this way the bottom half won't be ignored! Neat, if you want to use the lower left triangle for asymmetric coolers!

File: tests/test_compartments_saddle.py [hunk 2]
```diff
@@ -182,7 +182,7 @@ def test_get_digitized():
     df = pd.DataFrame(
         [["chr1", 0, 10, pd.NA]],
         columns=["chrom", "start", "end", "value"],
-    ).astype({"value": pd.Float64Dtype()})
+    ).astype({"value": pd.Int64Dtype()})
```
sergpolly, line 185: in the end - this should be Float64Dtype

gfudenberg: I thought int64dtype worked with newer version of pandas?

cooltools/snipping.py
```diff
@@ -361,11 +490,28 @@ def select(self, region1, region2):
...
-        self._isnan2 = np.isnan(self.clr.bins()["weight"].fetch(region2_coords).values)
+        if self.clr_weight_name:
+            self._isnan1 = np.isnan(
+                self.clr.bins()[self.clr_weight_name].fetch(region1_coords).values
+            )
+            self._isnan2 = np.isnan(
+                self.clr.bins()[self.clr_weight_name].fetch(region2_coords).values
+            )
+        else:
+            self._isnan1 = np.zeros_like(
```
sergpolly, line 501: mybe something like:
```python
l,h = clr.extent(region1_coords)
self._isnan1 = np.zeros(h-l)
l,h = clr.extent(region2_coords)
self._isnan2 = np.zeros(h-l)
```

maybe some tiny functions that tells how many bins are there in a region is needed (bioframe, cooler, cooltools ?)

sergpolly: not important at all, just a thought

cooltools/snipping.py
```diff
@@ -481,4 +643,209 @@ def snip(self, exp, region1, region2, tup):
...
+            features_df["lo"] = (features_df["start"] / clr.binsize).astype(int)
+            features_df["hi"] = (features_df["end"] / clr.binsize).astype(int)
+        else:
+            features_df["lo1"] = (features_df["start1"] / clr.binsize).astype(int)
+            features_df["hi1"] = (features_df["end1"] / clr.binsize).astype(int)
+            features_df["lo2"] = (features_df["start2"] / clr.binsize).astype(int)
+            features_df["hi2"] = (features_df["end2"] / clr.binsize).astype(int)
+
+    if view_df is None:
+        view_df = bioframe.make_viewframe(clr.chromsizes)
```
sergpolly, line 751: i'd say use `make_cooler_view` helper function - mostly for consitency with other tools
but also because make_viewframe(chromsizes) yields a view, where names are UCSC by default, which isn't so nice when working on full chromosomzes

cooltools/snipping.py
```diff
@@ -481,4 +643,209 @@ def snip(self, exp, region1, region2, tup):
...
+    else:
+        if not bioframe.is_contained(view_df, bioframe.make_viewframe(clr.chromsizes)):
```
sergpolly, line 753: coud use a is_view_compatible check, but we could replace it after

cooltools/snipping.py
```diff
@@ -481,4 +643,209 @@ def snip(self, exp, region1, region2, tup):
...
+            raise ValueError("view_df is out of the bounds of chromosomes in cooler.")
+
+    if min_diag == "auto" and clr_weight_name not in [None, False]:
```
sergpolly, line 756: clr_weight_name usage is not consistent with other tools - we should discuss and fix after we agree on it

sergpolly (commented):
looks great !
I think we could merge and start using, and then fix some inconsistencies after

tests/test_compartments_saddle.py
sergpolly: value should be float-like, nit int-like. It was me who changed it to Int64Dtype in the middle of my branch because of old pandas. Float64Dtype is newer than Int64Dtype
sorry for confusion

gfudenberg: ah, got it-- yes the value should be nan-able float before digitization, categorical or nan-able int after

sergpolly:
working on it now

requirements.txt
```diff
@@ -10,4 +10,4 @@ numpy
 pandas>=1.3
 scikit-learn
 scipy
-tables
```
sergpolly, line 13: why is this in a diff ? very strange

Issue #291, opened 2021-11-03
insulation updates for 0.5.0


golobor:
the whole set of insulation functions needs to be updated, e.g.

- [x] `diamond_insulation` does not have view_df as an input (currently takes chrms) https://github.com/open2c/cooltools/blob/938ec3007725d53312f757ebc51558e8293a5dcd/cooltools/cli/diamond_insulation.py#L55 

- [x] `find_boundaries` operates on chrms, not view regions. Should also preserve index of original dataframe, which would solve https://github.com/open2c/cooltools/issues/210

- [x] `calculate_insulation_score` operates on chrms, not view regions.https://github.com/open2c/cooltools/blob/938ec3007725d53312f757ebc51558e8293a5dcd/cooltools/insulation.py#L121
  - also https://github.com/open2c/cooltools/issues/266

- [x] in `insul_diamond`, allow `clr_weight_name` to be passed
currently, the weight column is hard-coded:
https://github.com/open2c/cooltools/blob/db9392a992fcad52c8d277f51e80ebb1a4c4e551/cooltools/insulation.py#L80

golobor:
thank you, @dmalzl!

gfudenberg:
for CLI:

- https://github.com/open2c/cooltools/issues/256

agalitsyna:
for CLI and calculate_insulation_score: 

- https://github.com/open2c/cooltools/issues/266

golobor:
we also need a single function that calculates scores, calls and filters boundaries (so called "user-facing" function as defined in #258 )

sergpolly:
@golobor 

just check insulation CLI code - there are literally just 2 function calls - and they looks user-facing enough to me:
```python
    # Calculate insulation score:
    ins_table = insulation.calculate_insulation_score(
        clr,
        view_df=view_df,
        window_bp=window,
        ignore_diags=ignore_diags,
        min_dist_bad_bin=min_dist_bad_bin,
        append_raw_scores=append_raw_scores,
        chunksize=chunksize,
        verbose=verbose,
    )

    # Find boundaries:
    ins_table = insulation.find_boundaries(
        ins_table,
        view_df=view_df,
        min_frac_valid_pixels=min_frac_valid_pixels,
        min_dist_bad_bin=min_dist_bad_bin,
    )
```

what else is needed ?

gfudenberg:
tests for insulation.insulation to be added post-release

Issue #272, opened 2021-09-07
cooler kwargs for cooltools functions


gfudenberg:
for clarity, pass 
```python
my_func_of_clr(
 clr, stuff, ..., clr_weight_name="weight")
```
 instead of `weight="weight"` or `weight_name="weight"`

agalitsyna:
Checked/added for:

- [ ] insulation
- [ ] snipping
- [ ] eigdecomp
- [ ] saddle
- [ ] coverage
- [ ] dotfinder
- [ ] expected

agalitsyna:
Looks like in some places we already use similar parameter, "balance", e.g. https://github.com/open2c/cooltools/blob/5ec504ffa57e7532ef7014d9ce683101644db9a2/cooltools/eigdecomp.py#L296
Here, "balance" is used both as the name of the weight column and as the flag for balancing (can be False). 

What I find surprising is that in the same function there is a separate usage of "weight" column, not related to the passed balancing weight: 
https://github.com/open2c/cooltools/blob/5ec504ffa57e7532ef7014d9ce683101644db9a2/cooltools/eigdecomp.py#L386 

These two roles are very different, and I suggest to use "balanced" as flag only and pass "clr_weight_name" as column name simultaneously. This requires some discussion and deciding what's the best for all.  

List of locations where the change is needed: 

- [ ] Insulation code, no separate "balanced" parameter at all:
https://github.com/open2c/cooltools/blob/5ec504ffa57e7532ef7014d9ce683101644db9a2/cooltools/insulation.py#L73
- [ ] Deprecated function for dense insulation (yet important for testing) has both balance and weight:
https://github.com/open2c/cooltools/blob/5ec504ffa57e7532ef7014d9ce683101644db9a2/cooltools/insulation.py#L391
- [ ] Saddles are fully transferred to clr_weight_name, but it is used for both indicator of balancing and retrieval of weights: 
https://github.com/open2c/cooltools/blob/5ec504ffa57e7532ef7014d9ce683101644db9a2/cooltools/saddle.py#L374
- [ ] dot finder code uses yet another parameter called "bal_v_name": 
https://github.com/open2c/cooltools/blob/5ec504ffa57e7532ef7014d9ce683101644db9a2/cooltools/dotfinder.py#L924
- [ ] expected code uses "weight_name" and "clr_weight_name" interchangeably: 
[...]

Issue #266, opened 2021-09-01
Insulation score does not ignore "NaN" regions; additionally, can diamond insulation be normalized to surrounding region?


gabedolpten:
Hi, there are two separate concerns here:
1. The insulation score does not seem to ignore regions that are NaN in the coolfile, leading to discrepancies
2. The diamond insulation score should be normalized to the regions proximal to the diamond, so that you are measuring "insulation compared to background". Is there some program which does this?

gfudenberg:
Thanks for the note & sorry for the slow reply.

It appears we pass:
-    min_frac_valid_pixels=0.66,
-    min_dist_bad_bin=0,
to `find_boundaries`, but should add this as an argument with sensible defaults to the `calculate_insulation_score`. (The current version of calculate_insulation_score assumed the user would use the returned table to filter insulation however they wish for downstream analysis).

The CLI should be updated to pass these to `calculate_insulation_score` as well.

We should decide, however, if we set min_dist_bad_bin=1 by default (it may be more interpretable) cc @golobor

gfudenberg:
is this addressed in new insulation function ? cc @Phlya @agalitsyna

agalitsyna:
insulation calculation of cooltools can now ignore NaNs and neighboring regions, controlled by min_dist_bad_bin param (introduced with pre-0.5.0 updates here https://github.com/open2c/cooltools/commit/98b11db82f799f120bc9a25026d92b1683297aba). 

@gfudenberg , what do you think about the normalization to the background?

gfudenberg:
Do you mean the `norm_by_median` argument in insul_diamond that is not
passed along to the user-facing function ?

agalitsyna:
Ah, okay, I had another explanation in mind. There is another widespread type of insulation normalization where the IS track is divided not by the whole chromosome average but by the moving centered average. This is not a traditional insulation score definition, and I am not sure who first implemented it. Some examples of "moving average insulation": [[1]](https://www.sciencedirect.com/science/article/pii/S0092867417303434#sec4) and [[2]](https://vaquerizaslab.github.io/fanc/fanc-executable/fanc-analyse-hic/domains.html). cooltools has the traditional insulation right now, following the original [Crane et al. 2016 work](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4498965/#S1title)

gfudenberg:
[...]

PR #299, opened 2021-11-06
Pre0.5.0

agalitsyna:
Insulation updates from https://github.com/open2c/cooltools/issues/291 (resolved).
Resolved with unchanged default min_dist_bad_bin=0: https://github.com/open2c/cooltools/issues/266 
Proposed solution to https://github.com/open2c/cooltools/issues/272

### DIFF
--- a/cooltools/expected.py
+++ b/cooltools/expected.py
@@ -15,8 +15,7 @@ from cooler.tools import partition
 import cooler
 import bioframe
 from .lib import assign_supports, numutils
-from .lib.common import is_compatible_viewframe, is_cooler_balanced
-from .lib.common import make_cooler_view, read_viewframe
+from .lib.common import is_compatible_viewframe, is_cooler_balanced, make_cooler_view
 
 where = np.flatnonzero
 concat = chain.from_iterable
--- a/cooltools/expected.py
+++ b/cooltools/expected.py
@@ -16,6 +16,7 @@ import cooler
 import bioframe
 from .lib import assign_supports, numutils
 from .lib.common import is_compatible_viewframe, is_cooler_balanced
+from .lib.common import make_cooler_view, read_viewframe
 
 where = np.flatnonzero
 concat = chain.from_iterable
--- a/cooltools/insulation.py
+++ b/cooltools/insulation.py
@@ -40,7 +40,7 @@ def get_n_pixels(bad_bin_mask, window=10, ignore_diags=2):
     return n_pixels
 
 
-def insul_diamond(pixel_query, bins, window=10, ignore_diags=2, norm_by_median=True):
+def insul_diamond(pixel_query, bins, window=10, ignore_diags=2, norm_by_median=True, clr_weight_name="weight"):
     """
     Calculates the insulation score of a Hi-C interaction matrix.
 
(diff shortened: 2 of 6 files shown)


cooltools/expected.py
```diff
@@ -16,6 +16,7 @@
 import bioframe
 from .lib import assign_supports, numutils
 from .lib.common import is_compatible_viewframe, is_cooler_balanced
+from .lib.common import make_cooler_view, read_viewframe
```
sergpolly, line 19: hm, was it working before that ? are we using make_cooler_view and read_viewframe here  ?

cooltools/insulation.py
```diff
@@ -70,16 +76,16 @@ def insul_diamond(pixel_query, bins, window=10, ignore_diags=2, norm_by_median=T
     sum_balanced = np.zeros(N)
```
sergpolly, line 77: should probably add that  `is_cooler_balanced` check here ...
I haven't propagated that check to many tools yet (should be everywhere basically) , but here is one example https://github.com/open2c/cooltools/blob/705f5d4e895aa4c16df6c210c7f1987a4754a978/cooltools/expected.py#L1459

for this case we could do some simple exception catching this:
```python
try:
   _ = is_cooler_balanced(clr, clr_weight_name, raise_errors=True)
except Exception as e:
    raise(f"cooler is not balanced or {clr_weight_name} is missing") from e
```

not sure it is useful, but it would be sort of trivial to allow `clr_weight_name=None` - i.e. insulation on unbalanced data as well . Not for v0.5 i think

cooltools/insulation.py
```diff
@@ -152,14 +170,26 @@ def calculate_insulation_score(
...
+                clr,
+                check_sorting=True,
+                raise_errors=True,
+            )
+        except Exception as e:
+            raise ValueError("view_df is not a valid viewframe or incompatible") from e
 
     bin_size = clr.info["bin-size"]
     ignore_diags = (
         ignore_diags
```
sergpolly, line 190: i think @Phlya was in favor of "simple" `if then else` instead of this one-liners - others agreed - should we modify ?

I also think a simple `if then else` is probably cleaner , unless a one-liner `if then else` really fits in 1 line

cooltools/insulation.py
```diff
@@ -181,41 +211,50 @@ def calculate_insulation_score(
...
+                    region_query, region_bins, window=win_bin, ignore_diags=ignore_diags, clr_weight_name=clr_weight_name
                 )
                 ins_track[ins_track == 0] = np.nan
                 ins_track = np.log2(ins_track)
 
             ins_track[~np.isfinite(ins_track)] = np.nan
 
-            ins_chrom["log2_insulation_score_{}".format(window_bp[j])] = ins_track
-            ins_chrom["n_valid_pixels_{}".format(window_bp[j])] = n_pixels
+            ins_region["log2_insulation_score_{}".format(window_bp[j])] = ins_track
```
sergpolly, line 244: replace with f-strings ? `f"log2...{window_bp[j]}"`

cooltools/snipping.py
```diff
@@ -6,6 +6,7 @@
 import bioframe
 
 from .lib.common import assign_regions, is_compatible_viewframe, is_compatible_expected
+from .lib.common import make_cooler_view, read_viewframe
```
sergpolly, line 9: was it working before ?! not sure how i missed it))

sergpolly (changes requested):
looks great ! let's merge after a couple of tiny edits

sergpolly:
I'll make changes myself - just checked out the branch

cooltools/insulation.py
sergpolly: moved this check to the user-facing function instead

cooltools/insulation.py
```diff
@@ -399,12 +471,16 @@ def _find_insulating_boundaries_dense(
...
-        ins_chrom = clr.bins().fetch(chrom)[["chrom", "start", "end"]]
-        is_bad_bin = np.isnan(clr.bins().fetch(chrom)["weight"].values)
+    ins_region_tables = []
+    for chrom, start, end, name in view_df[['chrom', 'start', 'end', 'name']].values:
+        region = [chrom, start, end]
+        ins_region = clr.bins().fetch(region)[["chrom", "start", "end"]]
+        is_bad_bin = np.isnan(clr.bins().fetch(region)[clr_weight_name].values)
 
-        m = clr.matrix(balance=balance).fetch(chrom)
+        if balance==True:
```
sergpolly: I've removed the balance option - to make API aligned with other tools that we have

sergpolly (approved):

PR #304, opened 2021-11-10
split expected cli into cis and trans

sergpolly:


cooltools/cli/cis_expected.py
```diff
@@ -35,39 +35,21 @@
...
+    " (no 4th column), UCSC-style region names are generated."
+    " Cis-expected is calculated for all chromosomes, when this is not specified."
+    " Note that '--regions' is the deprecated name of the option. Use '--view' instead.",
     type=click.Path(exists=True),
     required=False,
 )
-@click.option(
-    "--balance/--no-balance",
-    help="Apply balancing weights to data before calculating expected."
-    "Bins masked in the balancing weights are ignored from calcualtions.",
```
sergpolly, line 63: that's how you trigger raw expected caculations now:
```sh
cooltools cis-expected --clr-weight-name ""  CN.mm9.1000kb.cool
```
I don't like it much, but API is streamlined now ! no more `balance` and `clr_weight_name` at the same time

Issue #280, opened 2021-10-14
get_cis_expected API (smoothing, aggregation, etc)


gfudenberg:
- [ ] propagating NaNs through smoothing e.g. balanced.sum.smoothed should be NaN if balanced.sum is NaN.

- [ ] consider renaming balanced & areas here to something more interpretable (smoothed_values, smoothed_counts?)
https://github.com/open2c/cooltools/blob/f478c1c7b69d8dabeb383c9a5aab5d6d753a6417/cooltools/sandbox/expected_smoothing.py#L195

- [ ] its not currently clear what the agg argument is supposed to do-- it seems to deletes the count column (which also has slightly unexpected values)

![image](https://user-images.githubusercontent.com/19657573/137390804-880ac107-2175-4969-a0db-3b4a32e03c19.png)

cc @golobor

golobor:
thank you for these thoughtful comments!

> propagating NaNs through smoothing e.g. balanced.sum.smoothed should be NaN if balanced.sum is NaN.

not sure if I agree with this one! By definition, smoothing assumes that the "true" value of contact frequency for a given diagonal can be inferred from those from adjacent diagonals. If if some diagonal misses good pixels entirely, I would say, we should still be able to estimate its contact frequency, no?..

>  consider renaming balanced & areas here to something more interpretable (smoothed_values, smoothed_counts?)

Done!

>  its not currently clear what the agg argument is supposed to do-- it seems to deletes the count column (which also has slightly unexpected values)

Yes, that is a good point - I just pushed new API that does not have this argument and does not drop columns that were present in the cvd table: 
https://github.com/open2c/cooltools/blob/704f259c4df5bc1c61be915bf56c80b00a0298ed/cooltools/sandbox/expected_smoothing.py#L232
https://github.com/open2c/cooltools/blob/master/cooltools/sandbox/expected_smoothing_example.ipynb

gfudenberg:
going through the notebooks, noticed that the docstring for 
``` python
def get_cis_expected(
    ...
    intra_only=True,
    ...
):
 says intra_only=False returns all combinations, but it only returns non-symmetric combinations. 

Had we discussed a desired behavior?


https://github.com/open2c/cooltools/blob/ad018d6005b5df852cf39f6ec5321bdfab95106f/cooltools/expected.py#L984

sergpolly:
yes, - we've discussed ...
conclusion was to makr `intra_only=False` to return EVERYTHING - intra and inter together
[...]

PR #305, opened 2021-11-11
enable smooth and aggregate in cis-expected using Antons smoothing

sergpolly:
@golobor check this out - does this align with your vision ?
this is a bit quick and dirty , but Im not sure there is much more to it ...
I kept diffs as small as possible, to allow quicker review

if that looks ok - we can sandbox logbin


sergpolly:
have to admit - it was super easy ! - didn't not check how the results look like though

1 thing we should do is to prevent your function to dump _ALL_ of the additional columns ... maybe ...

golobor:
wow, looks good! i think it should do, for now - but could you also add an extra-flag for sigma_log10 , which defines the width of the smoothing kernel in the log10 space?

sergpolly:
@golobor  - what bout `window_sigma` ? it looks like it is also responsible for the width of smoothing ...

golobor:
oh, sorry, window_sigma is a technical parameter. the choice of the name is probably confusing - may be we can dedicate one of our post-release calls to this algorithm

Phlya:
We already dedicated this call to this algorithm :D See the comment in #280 for a suggested change to the API.

Issue #312, opened 2021-11-17
v0.5 roadmap


gfudenberg:
- [ ] set up tests/publish via Github actions!
- [ ] update track inputs to correspond to standard #271
- [ ] https://github.com/open2c/cooltools/issues/251  @Phlya @agalitsyna 
- [ ] Consider deprecating `cooltools.cli.genome` by creating bioframe CLI with these functions, or relying on cooler CLI? @nvictus #241 
- [ ] add a check for call compartments / cis_eig for the matching resolution of the phasing track and the cooler https://github.com/open2c/cooltools/issues/212
- [ ] split lib.common into core and sandbox, decide for each function where it goes
- [ ] #302
- [ ] tutorials for dots and apa
- [ ] document design choices, e.g.:
  - [ ] functions in core.numutils should not be imported into base namespace
  - [ ] pure numpy functions should be in core.numutils, even if used in one package (e.g. `ecdf()`) to make them more visible and used across modules.
  - [ ] cooltools functions should not rely on bioframe.contrib
  - [ ] main cooltools API modules and CLI should not depend on cooltools.contrib

PR #318, opened 2021-12-04
split-common-into-common-checks-io

gfudenberg:
addresses https://github.com/open2c/cooltools/issues/302 and several bullet points from https://github.com/open2c/cooltools/issues/312 refactoring

split `lib.common` into:
- lib.common (basically bioframe for cooltools),
- io (data download & reading from files),
- checks (checks!)

also:
- moved `mask_cooler_bad_bins()`, `view_from_track()` from api.saddle into lib.common
- moved `is_track()` from api.saddle into lib.checks
- removed underscore from `_is_expected`, since it seems equally useful to is_compatible_expected for a user

I wonder if the following functions should be renamed for more clarity:
- read_viewframe --> read_viewframe_from_file
- read_bedframe --> read_bedframe_from_file
- view_from_track --> make_view_from_track

Q: should we keep the from file io and the osf data download together or split?


cooltools/lib/checks.py
```diff
@@ -0,0 +1,502 @@
...
+
+    it0, it1 = tee(iterable) # duplicate the iterator
+    next(it1, None) # skip 1st element in "it1" copy
+    # check if all values in iterable are in ascending order
+    # similar to all(array[:-1] < array[1:])
+    _pairs_out_of_order = starmap(gt, zip(it0, it1) )
+    # no pairs out of order returns True, i.e. iterator is sorted
+    return not any(_pairs_out_of_order)
+
+def is_expected(
```
sergpolly, line 30: I'm not full convinced this one is useful for users - I had it not hidden at first, but then added `_` - because `is_compatible_expected(...verify_cooler=None...)` does roughly the same

I made it hidden to keep the total number of visible checks smaller, less confusing

what are the arguments for making it visible ?

sergpolly (commented):
looks great ! just two Qs:
1. make sure `mask_cooler_bad_bins` is not needed in `api.saddle` (where it used to be defined before)
2. are we sure `is_expected` needs to be visible ? it's pretty rudimentary and is part of `is_compatible_expected`

 - I'm for turning `read_{whatever}` to `read_{whatever}_from_file` !
 - I'm neutral to `iew_from_track --> make_view_from_track` change

cooltools/lib/checks.py
gfudenberg: Aha, hadn't realized that base case could be assayed w/o passing a view. Maybe rename to is_valid_expected? Compatible makes me think it should strictly require two inputs. I also think I saw the _is_expected used outside of is_compatible_expected.

cooltools/cli/saddle.py
```diff
@@ -286,7 +289,7 @@ def saddle(
     else:
         max_diag = -1
     if clr_weight_name:
-        track = api.saddle.mask_bad_bins(
```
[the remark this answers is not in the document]
gfudenberg: Tests ran so probably not! ?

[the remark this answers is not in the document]
sergpolly: yes, you're right - only used in CLI before digitization of the `track` ...

But now it makes me wonder if this step should be better incorporated into one of the user-facing functions ?
right now it is done in CLI outside of any user-facing functions `digitize` or `saddle` - it does not fit into `digitize` now, because `digitize` does not know about the `cooler` (now), but by the time we get to `api.saddle` it is too late ...

Does it "deserve" to be brought into `digitize` by making it depend on the `cooler` and/or some `bad_bin_mask` (optionally of course)?

Is masking "bad" bins important for saddle ?:
- masking bad bins in the track seems important and (?)could affect the result of the saddle(?) via outliers , maybe ...
- maybe should check with a track defined for every bin (e.g. GC-content) and test with and without masking bad bins from cooler

cooltools/lib/checks.py
sergpolly: it might have been used separately in the `read_expected_from_file` before, but now it shouldn't be used anywhere outside of `is_compatible_expected` (will check)

All for renaming !  `is_valid_expected` does make more sense indeed !
although I wanted to ask/discuss what do this check should return - shoud they return `True/False` and optionaly `raise`, or should they return `list` of "issues" with the input ... @nvictus had some suggestions - but in the rush before the release we kept them as is

cooltools/cli/saddle.py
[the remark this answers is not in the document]
sergpolly: should we make this Q a separate issue and proceed with the split ?

cooltools/lib/checks.py
sergpolly: oh yeah, `_is_expected` is still used separately in `read_expected_from_file` - this can (and should) be changed

cooltools/cli/saddle.py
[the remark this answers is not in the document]
gfudenberg: Yes let's make this a separate issue-- there may even be one about saddles that this can be added to

cooltools/lib/checks.py
gfudenberg: Ok! I'll:
- roll back to _is_expected and
- go ahead with the other renaming
- update read_expected_from_file to only depend on is_valid_expected

sergpolly: Sounds great !

gfudenberg: Running into an issue: it seems that is_compatible_expected will not accept None for the view. 

image

I guess I can just make this execute if verify_view is not None: https://github.com/open2c/cooltools/blob/6c9f4ff8ae8f8dade313812fe8c9391d773ddb5d/cooltools/lib/checks.py#L207
?

gfudenberg: also-- shouldn't this be moved inside the for loop? https://github.com/open2c/cooltools/blob/6c9f4ff8ae8f8dade313812fe8c9391d773ddb5d/cooltools/lib/checks.py#L237

cooltools/lib/checks.py
```diff
@@ -205,15 +212,8 @@ def _is_compatible_cis_expected(
...
-            view_name_col="name",
-        ):
-            raise ValueError(
-                "View regions are not in the expected table. Provide expected table for the same regions"
-            )
+        if verify_view is not None:
+            _is_expected_cataloged(expected_df, verify_view)
 
         # check if the number of diagonals is correct:
         if verify_cooler is not None:
```
sergpolly, line 219: could you move this inside of the `if verify_view is not None:` branch - as the `verify_cooler` related check relies on `verify_view`

cooltools/lib/checks.py
```diff
@@ -297,13 +297,9 @@ def _is_compatible_trans_expected(
...
-        _all_expected_regions = expected_df[["region1", "region2"]].values.flatten()
-        if not np.all(verify_view["name"].isin(_all_expected_regions)):
-            raise ValueError(
-                "View regions are not in the expected table. Provide expected table for the same regions"
-            )
+        if verify_cooler is not None:
+            _is_expected_cataloged(expected_df, verify_view)
 
-        # check if the number of diagonals is correct:
         if verify_cooler is not None:
```
sergpolly, line 303: same here - this should be inside if `if verify_view is not None`

cooltools/lib/checks.py
```diff
@@ -297,13 +297,9 @@ def _is_compatible_trans_expected(
 
         # Check that view regions are named as in expected table.
         # Check region names:
-        _all_expected_regions = expected_df[["region1", "region2"]].values.flatten()
-        if not np.all(verify_view["name"].isin(_all_expected_regions)):
-            raise ValueError(
-                "View regions are not in the expected table. Provide expected table for the same regions"
-            )
+        if verify_cooler is not None:
```
sergpolly, line 300: typo here - should be `verify_view` not `verify_cooler`

cooltools/lib/io.py
```diff
@@ -295,9 +294,12 @@ def read_expected(
 
     try:
         expected_df = pd.read_table(fname)
-        _ = is_expected(expected_df,
+        _ = is_valid_expected(
```
sergpolly, line 297: oh yeah - should be gone now, it's all covered by the second check

cooltools/lib/checks.py
sergpolly: oh yeah - it looks like `if n_diags_expected != n_diags_cooler: raise ...` should inside of the `for` loop over regions! indeed!
thank you !

cooltools/lib/checks.py
gfudenberg: thanks!

cooltools/lib/checks.py
gfudenberg: hmm-- should we try to create a view from the cooler if `verify_cooler` is passed but `verify_view` is None? or just leave as is?

gfudenberg: at the very least we'd need to update the docstring that the `verify_cooler` argument also requires a view!

cooltools/lib/io.py
```diff
@@ -254,7 +253,7 @@ def print_available_datasets(url=URL_DATA):
 """.format(**data))
 
 
-def read_expected(
+def read_expected_from_file(
```
gfudenberg, line 256: can we remove the basic input check, since this is done in `is_valid_expected()`? @sergpolly

sergpolly: ph, yeah - I think it should be fine.
Assuming `if verify_view is not None:` below is skipped

cooltools/lib/checks.py
sergpolly: > hmm-- should we try to create a view from the cooler if verify_cooler is passed but verify_view is None? or just leave as is?
yeah - probably this is a useful scenario indeed `is_valid_expected(expected_df,..., verify_view=None, verify_cooler=clr)`
I haven't thought about that before ...

gfudenberg:
ok, added the following to `_is_compatible_cis_expected` & `_is_compatible_trans_expected` 
```python
if verify_cooler is not None:
    verify_view = make_cooler_view(clr) if verify_view is None else verify_view
    ...
```

PR #322, opened 2022-01-13
Some support for cis/trans arithmetic

Phlya:
Coverage saves total cis contacts into info (when `store=True`).
Sample can have total cis contacts as target: tries to use the stored value and calculates it if it's not there.


cooltools/api/coverage.py
```diff
@@ -119,5 +119,5 @@ def coverage(
                     del grp["bins"][store_name]
                 h5opts = dict(compression="gzip", compression_opts=6)
                 grp["bins"].create_dataset(store_name, data=cov_arr, **h5opts)
-
+            grp.attrs.create("cis", np.sum(covs[0]), dtype=int)
```
sergpolly, line 122: just make sure it goes to cooler.info
otherwise it is super cool

cooltools/api/sample.py
```diff
@@ -71,6 +73,10 @@ def sample(
         The target sample size as a fraction of contacts in the original
         dataset. Mutually exclusive with `count`.
 
+    cis_target : bool
```
sergpolly, line 76: maybe rename to cis_count and move up to be close to the original `count`

cooltools/api/sample.py
```diff
@@ -88,7 +94,11 @@ def sample(
         clr = cooler.Cooler(clr)
 
     if count is not None and frac is None:
```
sergpolly, line 96:
```python
# mutually exclusive frac, count and cis_count
if frac is not None and count is None and cis_count is None:
    # work with frac
elif frac is None and count is not None and cis_count is None:
   # work with count
elif frac is None and count is None and cis_count is not None:
   # work with cis_count
else:
    raise
```

cooltools/api/sample.py
```diff
@@ -48,6 +49,7 @@ def sample(
     out_clr_path,
     count=None,
     frac=None,
+    cis_target=False,
     exact=False,
```
sergpolly, line 53: consider adding `ignore_diags` here ? or ultimately `dist_min` and `dist_max` ?
if it does not end up oin this PR - leave an issue,  maybe ? for the future

cooltools/api/coverage.py
Phlya: It does!

cooltools/api/sample.py
Phlya: Done!

cooltools/api/sample.py
Phlya: Thank you, done!

Phlya:
All done except ignore_diags, also added tests for coverage API, and fixed a bug in exact sampling (which was broken for me due to wrong dtype).

Phlya:
Apart from the idea about ignore_diags this is ready IMO

cooltools/api/sample.py
sergpolly: I'd say yes ... Also just make sure what `ignore_diags` means exactly - ignore diags to set the count target, of course - but what should happen with the ignored_diags in the subsamples matrix ? Should they be zero ?

Phlya: Yes meaning ... which of the options?
I'd say the calculation of the subsampling fraction should not take `ignore_diags` diags into account, but subsampling would be applied to all data including those diags

cooltools/api/sample.py
```diff
@@ -63,13 +65,21 @@ def sample(
...
+    cis_count : int
+        The target number of cis contacts in the sample.
+        Mutually exclusive with `count` and `frac`.
 
     frac : float
         The target sample size as a fraction of contacts in the original
-        dataset. Mutually exclusive with `count`.
+        dataset. Mutually exclusive with `count` and `cis_count`.
+
+    cis_target : bool
```
sergpolly, line 80: remove `cis_target`  from docstring ?

tests/test_sample.py
```diff
@@ -0,0 +1,59 @@
...
+        op.join(request.fspath.dirname, "data/CN.mm9.1000kb.test_sampled.cool")
+    )
+    # Test that deviation from expected total is very small
+    testing.assert_allclose(clr_result.info["sum"], 200000000, rtol=1e-3)
+
+
+# Exact sampling is very slow! So commented out
+# cooltools.api.sample.sample(
+#     clr,
+#     op.join(request.fspath.dirname, "data/CN.mm9.1000kb.test_sampled.cool"),
```
sergpolly, line 39: can we use a smaller input sample to uncomment this ?

Phlya: This is our normal test cooler at 1Mb resolution... So quite small! Idk we could add a 10Mb resolution one?

Phlya: Even with 10Mb resolution it takes a while... This whole file takes 95 sec to run through pytest on my laptop, which I guess is acceptable?

sergpolly: let's try this way - w'd need to optimize out tests at some point anyways

PR #324, opened 2022-01-20
Assign view paired

### LINKED ISSUE #251: deprecate cooltools copy of assign_regions, switch to bioframe version

Phlya:
Created a function to assign view to a bedpe-style dataframe, to be used in snipping and dotcaller. Still need to actually replace the deprecated function.
Aims to close #251


Phlya:
[Whoops accidentally based this on the branch for https://github.com/open2c/cooltools/pull/322...]

Phlya:
Tests fail for test_expected, and I haven't touched any of that...

gfudenberg:
My PR fixes that, but hasn't been merged in since numba iterative
correction isn't working (& hence eig_trans test)

cooltools/lib/common.py
```diff
@@ -9,6 +9,70 @@
...
+    view_name_col : str
+        Column of ``view_df`` with region names. Default 'name'.
+    drop_unassigned : bool
+        If True, drop intervals in df that do not overlap a region in the view.
+        Default False.
+    """
+    features = features.copy()
+    features.reset_index(inplace=True, drop=True)
+
+    bioframe.core.checks.is_bedframe(features, raise_errors=True, cols=cols[:3])
```
sergpolly, line 49: `col_left` or `cols1` and `cols_right`/`col2`

cooltools/lib/common.py
```diff
@@ -9,6 +9,70 @@
 import bioframe
 
 
+def assign_view_2D(
+    features,
+    view_df,
+    cols=["chrom1", "start1", "end1", "chrom2", "start2", "end2"],
```
sergpolly, line 15: split into `cols1` and `cols2` or `_left` `_right`

or keep them together but call differently `cols_paired`

Phlya:
Rename function to assign_view_paired.

cooltools/lib/common.py
```diff
@@ -9,6 +9,70 @@
...
+    cols_view=["chrom", "start", "end"],
+    features_view_cols=["region1", "region2"],
+    view_name_col="name",
+    drop_unassigned=False,
+):
+    """Assign region names from the view to each feature
```
sergpolly, line 21: add clarification that this function assigns the "regular"(for lack of a better work) view to each both sets of intervals `cols1` and `cols2`
and it does not work with a "proper" 2D view - as that one would need to be defined using a pair of intervals

cooltools/api/snipping.py
```diff
@@ -777,6 +801,13 @@ def pileup(
             .astype(int)
         )
     else:
+        # This is for the future, when doing trans pileups is possible (between regions)
+        # features_df["region1_offset"] = features_df["region1"].replace(
+        #     region_offsets_dict
```
sergpolly, line 806: remove ?

cooltools/api/snipping.py
```diff
@@ -777,6 +801,13 @@ def pileup(
             .astype(int)
         )
     else:
+        # This is for the future, when doing trans pileups is possible (between regions)
```
Phlya, line 804: remove

cooltools/api/snipping.py
```diff
@@ -713,8 +714,31 @@ def pileup(
...
+        features_df = assign_view_2D(
+            features_df,
+            view_df,
+            cols=["chrom1", "start1", "end1", "chrom2", "start2", "end2"],
+            cols_view=["chrom", "start", "end"],
+            features_view_cols=["region1", "region2"],
+            view_name_col="name",
+            drop_unassigned=False,
+        )
+        features_df["region"] = np.where(
```
Phlya, line 737: explain what this is for

sergpolly, line 737: leave a comment with an  explanation

cooltools/api/snipping.py
```diff
@@ -713,8 +714,31 @@ def pileup(
     else:
         raise ValueError("Unknown feature_df format")
 
-    features_df = assign_regions(features_df, view_df)
-    # TODO: switch to bioframe.assign_view upon update
+    if feature_type == "bed":
+        features_df = bioframe.assign_view(
+            features_df,
+            view_df,
+            cols=["chrom", "start", "end"],
```
Phlya, line 721: Remove parameters with hardcoded default values

cooltools/api/snipping.py
```diff
@@ -713,8 +714,31 @@ def pileup(
...
+            cols_view=["chrom", "start", "end"],
```
sergpolly, line 722: maybe hide these default params - and leaving a small comment about it (or not)

cooltools/lib/common.py
```diff
@@ -9,6 +9,76 @@
...
+        features,
+        view_df,
+        drop_unassigned=drop_unassigned,
+        df_view_col=features_view_cols[0],
+        view_name_col=view_name_col,
+        cols=cols_left,
+        cols_view=cols_view,
+    )
+    features[cols_right[1:]] = features[cols_right[1:]].astype(
+        int
```
sergpolly, line 68: can you check if this cast is needed for the final output - i.e. for after the second `assign_view` application ? just in case (also is it an expected behavior on bioframe side of things ? issue-worthy or not ?)

tests/data/CN.mm9.toy_regions.bed
```diff
@@ -1,2 +1,2 @@
 chr1	100000000	150000000	foo
-chr2	100000000	150000000	bar
+chr2	100000000	150000000	bar
```
sergpolly, line 2: does black do this ?
i'm not an expert on all things unix, but i think there should be a new line at the end e.g. https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline#:~:text=If%20content%20is%20added%20to,you%20actually%20wanted%20to%20see.

cooltools/lib/common.py
Phlya: This is because the df start off with numpy int dtypes, and then bioframe converts the other columns to numpy float64. But if it is pd.Int64Dtype from the beginning, this cast is not necessary.
In the second run of assign_view the cols_right[1:] get cast to pd.Int64Dtype, and the other ones are already that from this first cast... It's confusing. But the second cast is not necessary.
I don't know if this should be an issue in bioframe, I remember extended discussions about dtypes there and I think we couldn't find a way to avoid this sort of weird casting...

tests/data/CN.mm9.toy_regions.bed
Phlya: Maybe I did myself by accident...

cooltools/api/snipping.py
```diff
@@ -713,8 +696,20 @@ def pileup(
...
         raise ValueError("Unknown feature_df format")
 
-    features_df = assign_regions(features_df, view_df)
-    # TODO: switch to bioframe.assign_view upon update
+    if feature_type == "bed":
+        features_df = bioframe.assign_view(
+            features_df, view_df, df_view_col="region", view_name_col=view_name_col,
+        )
+    else:
+        features_df = assign_view_paired(features_df, view_df,)
```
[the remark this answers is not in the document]
Phlya: Good catch, thanks!

PR #325, opened 2022-02-01
fix-test_expected-after-bioframe-make_viewframe-name_style-update

gfudenberg:
changing the default behavior of bioframe.make_viewframe to not create UCSC names made the following code incompatible:
```python
common_regions = []
for i in range(4):
    chrom = chromosomes[i]
    halfway_chrom = int(chromsizes[chrom] / 2)
    # make halfway_chrom point "bin-aligned" according to anticipated binsize
    halfway_chrom = round(halfway_chrom / assumed_binsize) * assumed_binsize
    reg1 = (chrom, 0, halfway_chrom)
    reg2 = (chrom, halfway_chrom, chromsizes[chrom])
    common_regions.append(reg1)
    common_regions.append(reg2)

view_df = bioframe.make_viewframe(common_regions)
```

fixed by passing `name_style="ucsc"`

### DIFF
--- a/cooltools/api/eigdecomp.py
+++ b/cooltools/api/eigdecomp.py
@@ -471,9 +471,9 @@ def eigs_cis(
     # output table eigvec_table and eigvals_table
     for _region, _eigvals, _eigvecs in results:
         idx = bioframe.select(eigvec_table, _region).index
-        eigvec_table.at[idx, eigvec_columns] = _eigvecs.T
+        eigvec_table.loc[idx, eigvec_columns] = _eigvecs.T
         idx = bioframe.select(eigvals_table, _region).index
-        eigvals_table.at[idx, eigval_columns] = _eigvals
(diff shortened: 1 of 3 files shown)


gfudenberg:
figured out one reason why expected checks failed

older version
image

v1.4.0 pre fix
image

basically the multi-index is dropped in cases when there is a single group left after `groupby`.  

fixed by adding a `set_index(groupby)` before the groupby in `_agg_smooth_cvd`

this was hard to debug-- some tests for intermediate parts of this smoothing would be helpful!

gfudenberg:
still getting a strange error with eigs_trans-- 
``` 
<Result SystemError('CPUDispatcher(<function iterative_correction_symmetric at 0x7f90cb278550>) returned a result with an error set')>.exit_code```

it sounds like this might be numba-related... any ideas @golobor @nvictus ?

PR #327, opened 2022-02-09
Dots cleanup and high-level API

sergpolly:


cooltools/api/dotfinder.py
```diff
@@ -62,16 +81,26 @@ def recommend_kernel_params(binsize):
...
     elif binsize >= 4000:
-        # ~5 kb
+        # ~ 5 kb
         w, p = 7, 4
     else:
         # < 5 kb - is probably too fine ...
         raise ValueError(
             f"Provided cooler has resolution {binsize} bases,"
-            " which is too fine for analysis."
+            " which is too fine for analysis. Provide custom kernels"
```
gfudenberg, line 90: "which is too fine for automated kernel recommendation."

cooltools/api/dotfinder.py
```diff
@@ -31,11 +38,21 @@
...
+    recommended for dot-calling based on the resolution
+    of the input data.
+
+    At the moment it recommends standard 4-kernels originally
+    defined for the HiCCUPS method: donut, horizontal, vertical,
+    lowerleft
+
+    Kernels are recommended for resolutions near 5, 10
+    and 25 kb at the moment. Typically "dots" are not visible
+    at lower resolutions (>28kb) and data is too sparse for
```
gfudenberg, line 53: most datasets are too sparse

gfudenberg: maybe there are kernel params we can recommend for 1kb? cc @nvictus @itsameerkat

cooltools/api/dotfinder.py
```diff
@@ -44,16 +61,18 @@ def recommend_kernel_params(binsize):
...
-    # TODO: rename w, p to wid, pix probably, or _w, _p to avoid naming conflicts
+    # define "donut" parameters w and p based on the resolution of the data
+    # w is the outside width, and p is the internal width.
     if binsize > 28000:
         # > 30 kb - is probably too much ...
         raise ValueError(
-            f"Provided cooler has resolution {binsize} bases,"
-            " which is too coarse for analysis."
+            f"Provided cooler has resolution of {binsize} bases,"
+            " which is too coarse for analysis. Provide custom kernels"
```
gfudenberg, line 74: too coarse for automated analysis

or 
too coarse for automated kernel recommendation

cooltools/api/dotfinder.py
```diff
@@ -62,16 +81,26 @@ def recommend_kernel_params(binsize):
         # ~ 10 kb
         w, p = 5, 2
     elif binsize >= 4000:
-        # ~5 kb
+        # ~ 5 kb
```
gfudenberg, line 84: I think the valueError is clear enough so these comments can be removed

gfudenberg:
Given this module is complicated enough, I wonder if it would be useful to have a module-level docstring that explains the user-facing function and the different steps performed, as well as what the defaults support. E.g. copying the commented steps in the user-facing function to the top and mentioning the functions called in each step (https://github.com/open2c/cooltools/blob/aaa2125bca814491d30f1510434d82161974095e/cooltools/api/dotfinder.py#L1661 etc.). Another option could be to put the overall workflow in the Notes section of `dots()`.

cooltools/api/dotfinder.py
```diff
@@ -746,9 +676,8 @@ def score_tile(
     exp_v_name,
```
gfudenberg, line 676: I think expected_val_col is used elsewhere in cooltools?

cooltools/api/dotfinder.py
```diff
@@ -1275,16 +1148,16 @@ def clustering_step(
...
-    ######################################################################
+def cluster_filtering_hiccups(
+    centroids,
+    obs_raw_name=observed_count_name,
+    enrichment_factor_1=1.5,
+    enrichment_factor_2=1.75,
+    enrichment_factor_3=2.0,
+    FDR_orphan_threshold=0.02,
+):
+    # ad hoc filtering by enrichment, FDR for singletons etc.
```
gfudenberg, line 1159: ad hoc filtering of centroids by enrichment relative to recommended kernels, 

(just thinking to note somehow that this depends on the recommend_kernels)

gfudenberg:
it might be clearer to pass tile_cij to get_adjusted_expected_tile_some_nans() rather than introduce this variable origin (or rename to something keeping tile in the name, e.g. tile_start_ij) https://github.com/open2c/cooltools/blob/aaa2125bca814491d30f1510434d82161974095e/cooltools/api/dotfinder.py#L726

cooltools/api/dotfinder.py
```diff
@@ -1569,8 +1416,7 @@ def scoring_and_extraction_step(
         verbose=very_verbose,
     )
 
-    # composing/piping scoring and histogramming
-    # together :
+    # composing/piping scoring and histogramming together :
     job = lambda tile: to_extract(to_score(tile))
```
gfudenberg, line 1420: do I understand correcdtly that the hists are just per tile?

gfudenberg:
dots() could be more readble if all the kernel preparations were refactored into a separate function ie. starting from here https://github.com/open2c/cooltools/blob/aaa2125bca814491d30f1510434d82161974095e/cooltools/api/dotfinder.py#L1590 to tiles

gfudenberg:
should scoring_step be removed if only scoring_and_histogramming_step is used?

cooltools/api/dotfinder.py
sergpolly: updated text. we'll investigate 1-2kb resolution

sergpolly:
yep  - removed `scoring_step` it was nice to have for demonstration purposes - i.e. calculate those "donut"-based enrichments for every pixel and return them as a pixel-table. But it is easy to write such function in any notebook if needed - so removing it from the main codebase

cooltools/api/dotfinder.py
sergpolly: yeah - because bins are pre-defined (lambda-chunks) we just built "partial"-histograms per tile and then aggregate them together. We cannot have a shared global histogram for several processes to write to - because - Python ...

cooltools/api/dotfinder.py
sergpolly: added doctstring to this and renamed vars

sergpolly:
almost done and ready for review
TODO:
 - add description of dot-calling steps in the user-friendly API function `dots`
 - notebook in examples ...

address next time - kernels handling !

sergpolly:
this one is ready to go:
 - major refactoring of dotfinder +  user facing dots function
 - fix for the new numba issue is here (numba doesn't handle dicts well, so dict -> tuple fix applied with docstring)
 - addressed several deprecation warnings: np.str -> np.str_; pd.append deprecations, scipy namespaces warnings, etc

@gfudenberg I've addressed all your suggestions - do you want to take a look at it one more time ? Otherwise I would just merge it to speed things up a bit

gfudenberg:
wow, thanks for this update @sergpolly, this looks great !!!

I tried to give an update to some docstrings, feel free to re-modify/let me know if I made things less rather than more clear. I'm in favor of merging as well!

Here are a few more questions/comments/notes, that could maybe be quickly added if you agree (e.g. first point) or made into issues if you think its worth it:

- for CLI, maybe mention that the expected column specification is optional & provide an example? E.g. if the following is correct: ```
The path to a tsv-like file with a compatible expected table. Use the '::' syntax to optionally specify a column name (e.g. '/path/to/expected::balance.avg').```

- mention in the dots() docstring that the cluster_filtering can in the future be allowed to pass a user-defined function?

- slightly confusing to refer to both lamba bins and lamba chunks in `dots` docstrings. can we go with bins everywhere?

- it's not clear why `columns_for_output` is only specified in `cluster_filtering_hiccups()`. Maybe move this outside that function?

- the reference to deduplication is a little confusing https://github.com/open2c/cooltools/blob/2c17e395a4e252cf78ee9127d1de1d1b49e0e475/cooltools/api/dotfinder.py#L593 since it seems later that significant pixels should not be duplicated: https://github.com/open2c/cooltools/blob/2c17e395a4e252cf78ee9127d1de1d1b49e0e475/cooltools/api/dotfinder.py#L1394.

- it seems like dots passes view regions rather than expected regions, could renaming the argument make it more clear? (e.g. view_regions or just regions) https://github.com/open2c/cooltools/blob/2c17e395a4e252cf78ee9127d1de1d1b49e0e475/cooltools/api/dotfinder.py#L1024. Its also not clear what the 'comply_fdr' column refers to, can that be deleted from docstring?

- should we note that the `cluster_filtering_hiccups()` is only compatible when four kernels with names matching the pre-specified names have been passed (and raise an error if they do not)?

- should we update the `read_expected_from_file()` docstring to point to the schema https://github.com/open2c/cooltools/blob/2c17e395a4e252cf78ee9127d1de1d1b49e0e475/cooltools/lib/io.py#L290, and copy an up-to-date example from the issue to schemas?

p.s. how did you figure out the the numba fix? fixing that is awesome as well !

sergpolly:
Nice ! thank you @gfudenberg !
i like docstring updates!

> for CLI, maybe mention that the expected column specification is optional & provide an example? E.g. if the following is correct: The path to a tsv-like file with a compatible expected table. Use the '::' syntax to optionally specify a column name (e.g. '/path/to/expected::balance.avg').

turned into an issue, as it affects several CLIs

> mention in the dots() docstring that the cluster_filtering can in the future be allowed to pass a user-defined function?

not sure yet - let's not over - promise. Maybe should make it more clear that significant pixels can be used as-is and/or for custom clustering

> should we update the read_expected_from_file() docstring to point to the schema, and copy an up-to-date example from the issue to schemas?

turned into a separate issue

> how did you figure out the the numba fix? fixing that is awesome as well !

the error was saying something about `Dict[something[something]]` - so I just looked it up - numba and python dicts, turns out only types dicts are supported now - i.e. {str:float, str:float, str:float, ...}  - where all keys are of the same type and all values are of the same type. And we had {str:int, str:bool} - which was crashing (error message was unreadable !)... So i replaced that dict with a tuples (int,bool) and updated the docstring . I think it is internal enough function - not to worry about slight API change

TODO:
 - > slightly confusing to refer to both lamba bins and lamba chunks in dots docstrings. can we go with bins everywhere?
 - > it's not clear why columns_for_output is only specified in cluster_filtering_hiccups(). Maybe move this outside that function?
 - > the reference to deduplication is a little confusing
 - > Its also not clear what the 'comply_fdr' column refers to, can that be deleted from docstring?
 - > should we note that the cluster_filtering_hiccups() is only compatible when four kernels with names matching the pre-specified names have been passed (and raise an error if they do not)?
 - `expected_regions` in clustering step - rename to something else

will finish TODOs when I can

Issue #335, opened 2022-03-04
Potential bug in filling in NaNs in snipping


Phlya:
I think the way we (or I?) implemented filling in NaNs in snipping can cause wrong assignment of NaNs with multiprocessing, or when multiple selections have been made before snipping.
https://github.com/open2c/cooltools/blob/c610f43be909ecf89f561f76a211768e00229550/cooltools/api/snipping.py#L353
This block would overwrite the self._isnan1/2 arrays for each selection, while it should be stored in a dict for each region, perhaps.

agalitsyna:
Yeah, looks like a potential problem. There are two maps: one in [pileup_legacy](https://github.com/open2c/cooltools/blob/c610f43be909ecf89f561f76a211768e00229550/cooltools/api/snipping.py#L241)  and one in [_pileup](https://github.com/open2c/cooltools/blob/c610f43be909ecf89f561f76a211768e00229550/cooltools/api/snipping.py#L201). The outer parallelized map is propagated to the first function, but not the second. Thus, regions are processed in parallel, but the individual snippets are cut out sequentially.

What if you simply change the order of them? I mean, you can propagate `map` parameter to the _pileup and use inside while keeping the processing of regions sequential, which will guarantee that the snipper object attributes are not changed on the fly? Have you tried it?

Phlya:
I have not. If that approach has the same speed, it would be better because it won't require more memory to use multiple cores!
But I think we should fix this issue independently of the maps: in case someone wants to use the snippers directly...

PR #339, opened 2022-03-10
Fix snipping nan filling for multiprocessing

Phlya:
Addressing https://github.com/open2c/cooltools/issues/335, for now in a straightforward way of storing locations of bad bins in a dict per region

### DIFF
--- a/cooltools/api/snipping.py
+++ b/cooltools/api/snipping.py
@@ -82,12 +82,7 @@ def expand_align_features(features_df, flank, resolution, format="bed"):
 
 
 def make_bin_aligned_windows(
-    binsize,
-    chroms,
-    centers_bp,
-    flank_bp=0,
-    region_start_bp=0,
-    ignore_index=False,
+    binsize, chroms, centers_bp, flank_bp=0, region_start_bp=0, ignore_index=False,
 ):
     """
     Convert genomic loci into bin spans on a fixed bin-segmentation of a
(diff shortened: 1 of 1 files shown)


cooltools/api/snipping.py
```diff
@@ -305,10 +300,7 @@ def __init__(self, clr, cooler_opts=None, view_df=None, min_diag=2):
             # Make sure view_df is a proper viewframe
             try:
                 _ = is_compatible_viewframe(
-                    view_df,
-                    clr,
-                    check_sorting=True,
-                    raise_errors=True,
+                    view_df, clr, check_sorting=True, raise_errors=True,
```
sergpolly, line 303: thank you , black !

cooltools/api/snipping.py
```diff
@@ -351,19 +344,23 @@ def select(self, region1, region2):
             region1_coords, region2_coords
```
sergpolly, line 344: could you please please leave a comment here that matrix is fetched sparse here ?

sergpolly:
actually we have to double check if this is an issue at all ...
depending on how multiprocessing is used - it might actually create as many copies of snipper-instance as there are processes and nothing would overwrite the parent-objewct-instance ...

we just have to double check the initial premise of this 

one argument against is that if different chile-processes were indeed overwriting the parent `self._isnan` atribute - that something should have broke - just because regions have different length

Phlya:
See [discussion](https://github.com/open2c/cooltools/discussions/379)

PR #340, opened 2022-03-14
Docstrings for snippers

Phlya:
Start docs for low-level snipping by creating docstrings for snippers

### DIFF
--- a/cooltools/api/snipping.py
+++ b/cooltools/api/snipping.py
@@ -82,12 +82,7 @@ def expand_align_features(features_df, flank, resolution, format="bed"):
 
 
 def make_bin_aligned_windows(
-    binsize,
-    chroms,
-    centers_bp,
-    flank_bp=0,
-    region_start_bp=0,
-    ignore_index=False,
+    binsize, chroms, centers_bp, flank_bp=0, region_start_bp=0, ignore_index=False,
 ):
     """
     Convert genomic loci into bin spans on a fixed bin-segmentation of a
@@ -203,7 +198,7 @@ def _pileup(data_select, data_snip, arg):
     return np.dstack(stack), feature_group["_rank"].values
 
 
-def pileup_legacy(features, data_select, data_snip, map=map):
+def select_snip(features, data_select, data_snip, map=map):
     """
     Handles on-diagonal and off-diagonal cases.
 
@@ -222,7 +217,9 @@ def pileup_legacy(features, data_select, data_snip, map=map):
     data_snip : callable
         Callable that takes data, mask and a 2D bin span (lo1, hi1, lo2, hi2)
         and returns a snippet from the selected support region
-
+    
+    map : callable
+        Callable that works like builtin `map`.
 
     """
     if features["region"].isnull().any():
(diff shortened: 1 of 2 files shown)


Phlya:
Also renamed the `pileup_legacy` function to `select_snip`, since that's what it does (selects and snips). But open for other ideas for that, it's a bit of a weird one to name IMO.

cooltools/api/snipping.py
```diff
@@ -82,12 +82,7 @@ def expand_align_features(features_df, flank, resolution, format="bed"):
```
sergpolly, line 82: maybe add a module-wide docstring explaining basic technical details and caveats of snipping
 - what are the basic steps?
 - why pre-fetch entire region-wide matrix instead of fetching individual dense snips
 - are there any resource-related restrictions ? RAM ?

cooltools/api/snipping.py
```diff
@@ -203,7 +198,7 @@ def _pileup(data_select, data_snip, arg):
     return np.dstack(stack), feature_group["_rank"].values
 
 
-def pileup_legacy(features, data_select, data_snip, map=map):
+def select_snip(features, data_select, data_snip, map=map):
```
sergpolly, line 201: looks like an internal-enough function to become `_select_snip` at the very least

sergpolly: can we call it `_pileup` - why not ?

sergpolly: hide existing `_pileup` inside of this function, and rename that internal func to `_extract_pileup`/`_create_stack`/`extract_Stack` ...

and then reuse the name `_pileup` for *this* function (currently `select_snip`)
@agalitsyna what do you think ?

tests/test_snipping.py
```diff
@@ -180,7 +180,7 @@ def test_pileup(request):
     # assert stack.shape == (5, 5, 2)
 
 
-def test_ondiag_pileup_legacy_with_expected(request):
+def test_ondiag_select_snip_with_expected(request):
```
sergpolly, line 183: don't forget to change the name if we rename `select_snip`

gfudenberg:
was this branch still active btw?

Phlya:
Mh been a while - looks like I didn't address the comments about renaming some internal/legacy functions...

cooltools/api/snipping.py
Phlya: I don't want to put a function inside a function tbh, but renaming is great, thanks.

Phlya:
How do you like this? @sergpolly

cooltools/api/snipping.py
agalitsyna: renaming is great, I keep using this function as if it's not internal

cooltools/api/snipping.py
```diff
@@ -339,6 +317,23 @@ def __init__(self, clr, cooler_opts=None, view_df=None, min_diag=2):
         self.min_diag = min_diag
 
     def select(self, region1, region2):
+        """Select a portion of the cooler for snipping based on two regions in the view
```
gfudenberg, line 320: very helpful !

Phlya:
TODO:
- Move `pileup` docstring to the top of the module (see dotfinder for an example).
- Add explanation about assigning regions to features
- Mention expanding the features by `flank` and aligning to the cooler bins
- How are these prepared features used in `_pileup`?

PR #342, opened 2022-03-22
small updates to dots, regression fix

sergpolly:
- some minor usability updates
 - fixed performance regression when extracting enriched pixels from scored_df using threshold by lambda-bins

### DIFF
--- a/cooltools/api/dotfinder.py
+++ b/cooltools/api/dotfinder.py
@@ -73,6 +73,7 @@ from functools import partial, reduce
 import multiprocess as mp
 import logging
 import warnings
+import time
 
 from scipy.linalg import toeplitz
 from scipy.ndimage import convolve
@@ -93,6 +94,7 @@ from ..lib.checks import (
 
 from bioframe import make_viewframe
 
+warnings.simplefilter(action="ignore", category=pd.errors.PerformanceWarning)
 logging.basicConfig(level=logging.INFO)
 
 # this is to mitigate and parameterize the obs.raw vs count controversy:
@@ -988,6 +990,8 @@ def determine_thresholds(gw_hist, fdr):
         # run cumulative min, on the array of adjusted p-values
         # to make sure q-values are monotonously decreasing with pvals
         qvalues[k] = qvalues[k].mask(qvalues[k] > 1.0, 1.0)
+        # cast categorical index of dtype-interval to proper interval index
+        threshold_df[k].index = pd.IntervalIndex(threshold_df[k].index)
 
     return threshold_df, qvalues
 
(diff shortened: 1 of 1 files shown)


cooltools/api/dotfinder.py
```diff
@@ -1020,9 +1024,12 @@ def extract_scored_pixels(scored_df, thresholds, obs_raw_name=observed_count_nam
     for kernel_name, threshold in thresholds.items():
         # locally adjusted expected (lambda) of the scored pixels:
         lambda_of_pixels = scored_df[f"la_exp.{kernel_name}.value"]
-        # extract lambda_threshold for every scored pixel based on its estimated lambda, using
-        # interval-indexed 'threshold' to query it by the lambda-values of each scored pixel
-        threshold_of_pixels = threshold.loc[lambda_of_pixels]
+        # reconstruct edges of lambda bins from threshold's index:
+        ledges_reconstruct = np.r_[threshold.index.left, threshold.index.right[-1]]
+        # find indices of lambda-bins where pixels belong
+        lbin_idx = pd.cut(lambda_of_pixels, ledges_reconstruct, labels=False)
```
sergpolly, line 1030: one could simply do 
```python
threshold_of_pixels = threshold.loc[lambda_of_pixels]
```
but it is ~20 times slower, so - had to rethink it

cooltools/api/dotfinder.py
```diff
@@ -1663,7 +1682,8 @@ def dots(
     output_cols += [f"la_exp.{k}.qval" for k in kernels]
 
     # 4b. filter by enrichment and qval
-    if cluster_filtering is None:  # default - engage HiCCUPS filtering
+    if (cluster_filtering is None) or cluster_filtering:
```
sergpolly, line 1685: cluster_filtering=True makes sense ! let it be another option - it is confusing otherwise

so, now - `None` and `True` enable this filtering, while `False` deactivates it

cooltools/api/dotfinder.py
sergpolly: #343 explains this in detail

cooltools/api/dotfinder.py
sergpolly: @gfudenberg could you look into this `cluster_filtering` flag - we've discussed it together - but last time I used it , it seemed very confusing to me, so I changed its behavior ...

You were suggesting `cluster_filtering=None` - to be default and engage filtering, and `cluster_filtering=False` to switch filtering off. By mistake tried `cluster_filtering=True` and it all crashed in a non-intuitive way

i just think it is worth keeping `cluster_filtering=True` functional

PR #348, opened 2022-03-31
bins = clr.bins()[:] raises "ValueError: could not broadcast input ar…

adadiehl:
…ray from shape (N,) into shape (N-1,)" when used on cooler matrices from certain software. Replace with bins = clr.bins(); bins = bins[:len(bins)-1] to avoid error.

### DIFF
--- a/cooltools/api/expected.py
+++ b/cooltools/api/expected.py
@@ -316,7 +316,8 @@ def make_diag_tables(clr, regions, regions2=None, clr_weight_name="weight"):
             ]
         ).to_numpy()
 
-    bins = clr.bins()[:]
+    bins = clr.bins()
+    bins = bins[:len(bins)-1]
     if clr_weight_name is None:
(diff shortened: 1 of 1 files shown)


adadiehl:
For now, please ignore this request. After further testing, this hack does not work for all cases as the mismatch between the length of the bins object and apparent number of bins being variable at different resolutions. Agreed, it is odd that the canonical way does not work. The matrices I'm working with are from this workflow: https://github.com/nanoporetech/Pore-C-Snakemake, with the main repository here: https://github.com/nanoporetech/pore-c. Seems like the problem is likely in cooler construction. Will leave this open for now while I look into it further.

adadiehl:
Closing this. Further investigation verifies that this error tracks to problems with the cooler matrices from the nanopore pipeline.

adadiehl:
Closing.

Issue #271, opened 2021-09-07
cooler-aligned genomic tracks: definitions & assumptions 


gfudenberg:
for now: 
- should be dataframes with 4 columns
- interpret them as ['chrom','start','end','value']

used in functions as the following:
`my_func_of_track(track, kwargs*)`

this overwrites the old tuple (dataframe, value_column) format:
`my_func_of_track_OLD( (track, 'track_val'), kwargs*)`

for discussion: which scenarios cause issues for tracks?
- missing bins  
- not contained
- track bins that are nans in the cooler.bins()['weight']
- bins do not match track (offset or resolution or trimming ends or 0/1...)
- extra chromosome in track or cooler

golobor:
we would also need cooltools.core.checks.is_cooler_aligned(track, clr)

gfudenberg:
addressed in #320

gfudenberg:
addresses multiple points in 0.6 roadmap #312

Issue #180, opened 2020-08-25
Add a check to ensure reference track and cooler have the same resolution


cmdoret:
Hello, 

I just wanted to report what I think to be a bug in the CLI. When I use the option `--reference-track`, the compartments end up all positive and seem wrong (no correlation with gene expression). When I don't use the `--reference-track` option, the eigen vectors are not flipped / ranked, but they are fine.

When using the API directly via `cooler_cis_eig`, everything works well even when I provide a phasing track.

I encountered this issue with 2 separate dataset (mouse and human). Here is the head of the (tab-separated) reference-track file I used:

```
chrom   start   end     rna_coverage
chr1    0       10000   654.3748188
chr1    10000   20000   0
chr1    20000   30000   0
chr1    30000   40000   1445.995385
chr1    40000   50000   1418.11489
```

Versions:
* python: 3.7.8
* cooltools: 0.3.2
* pandas: 1.1.0
* numpy: 1.19.1

I'll be having a look at the code, but if you have any idea, that'd be great :)
Best

cmdoret:
OK I think I figured it out !
It's because the resolution of my cool file and the interval size on the bedgraph was different.
The problem happens during the merge of the phasing and bins tables:
https://github.com/mirnylab/cooltools/blob/95e744f43581b62c8baba3bc3b74ceff13f8cdf3/cooltools/cli/call_compartments.py#L131

I would suggest to either raise an error if resolutions are different, or to merge only on chrom and start with a warning, maybe ?

gfudenberg:
proposed function: `checks.cooler_bins_in_track()`
-- for each bin in the cooler bintable, check that it is present in the track
-- with possible exception of last bin on each chromosome

gfudenberg:
update 12/08/2021:

requirements above may be too strict, and current implementations in saddle & eigs are not compatible. what we want is also not really a check, since it returns a modified track.

new proposal 
```python
lib.common.align_track_to_clr(track,
                                                    clr,
                                                    view_df=None,
                                                    clr_weight_name='weight',
                                                    mask_bad_bins=True,
):
```

- check lib.checks.is_track() (no overlaps, bedframe, numeric 4th column)
- check median track bin size, cooler bin size. raise error if disagreeing
- try to pd.merge on ['chrom', 'start']
  - raise warning if <50% of valid clr.bins are assigned a value
  - raise an error if no values assigned
[...]

PR #360, opened 2022-05-26
BUGFIX: changed raise ValueError in align_track_with_cooler to just a…

kannandeepti:
… warning

Previously, the align_track_with_cooler() function raised a Value
Error if the track values of a given region in the viewframe were
all NaN. However, this check is not necessary because NaN values
are anyways already masked.

### DIFF
--- a/cooltools/lib/common.py
+++ b/cooltools/lib/common.py
@@ -368,7 +368,7 @@ def align_track_with_cooler(
     num_valid_bins = valid_bins.sum()
     num_assigned_bins = (clr_track["value"][valid_bins].notna()).sum()
     if num_assigned_bins == 0:
-        raise ValueError("no track values assigned to cooler bintable")
+        warnings.warn("no track values assigned to cooler bintable")
     elif num_assigned_bins < 0.5 * np.sum(valid_bins):
         warnings.warn("less than 50% of valid bins have been assigned a value")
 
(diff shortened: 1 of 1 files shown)


agalitsyna:
Hey, @gfudenberg. Is there a reason why empty regions in align track with cooler give an error rather than a warning? (Authorship deduced from git blame)

I'm thinking about Deepti's case: why would one chromosome having all NaNs in the dataframe before assignment constitute an issue in the downstream analysis?

gfudenberg:
Hi-- I don't know if I quite understand Deepti's case, but I think we discussed this in 
https://github.com/open2c/cooltools/issues/271
https://github.com/open2c/cooltools/issues/180#issuecomment-990391283
and the idea was that you should pass a view that does not have the offending chromosome, since it is one line. It may have helped streamline assumptions in some downstream applications as well (saddles?)

agalitsyna:
Thanks for getting back to it! I think the non-trivial thing here is that align_track_with_cooler does not distinguish absent bins in the track and NaN values in the track, and it's not entirely clear why. 
Imagine that some chromosome has all NaNs in the track (e.g. some mark is completely absent on this chromosome). It's expected that the bins from this chromosome do not contribute to aggregation, but why would this cause an error? 

The solution might be to raise an error only if the chromosome is completely absent form the track (or some bins of it, as you propose in https://github.com/open2c/cooltools/issues/271).

gfudenberg:
Maybe it is the use in functions like cis_eig? E.g. if there is this all NaN chromosome, maybe this would throw an error when you try to correlate?
https://github.com/open2c/cooltools/blob/45a56dfd07b8058b908763cc7c8e7db3fc52875e/cooltools/api/eigdecomp.py#L435

agalitsyna:
Okay, these are all good points that we discuss here. I introduced the parameter for ignoring missing values in alignment, and add it as a control option for saddle (because the chromosomes with all NaNs do not really affect the output). For eigdecomp you are right, @gfudenberg , there should be at least some value per each chromosome, otherwise phasing does not make sense. I set this new parameter to default True in eigdecomp.

To sum up, this is small update of align_track_with_cooler, but might be important to those practicing various types of saddles.

PR #367, opened 2022-06-28
Multithread insulation

agalitsyna:
Insulation API and CLI was not parallel for some reason. I made it parallel and it seems to speed up computations. 

Here is an example from open2c_examples insulation notebook: 

image

One thing that I do not understand completely is behavior of CSRSelector in multithreading mode. I had to create multiple instances of selectors for each thread. Are multiple CSRSelectors  guaranteed to work in order on a single cooler in multithreading? @nvictus

### DIFF
--- a/cooltools/api/insulation.py
+++ b/cooltools/api/insulation.py
@@ -1,5 +1,6 @@
 import re
 import logging
+
 logging.basicConfig(level=logging.INFO)
 
 import warnings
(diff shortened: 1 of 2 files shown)


nvictus:
The `CSRSelector` does carry around read-only open file state which can cause issues in multiprocessing contexts when child processes use inherited file handles from the parent process. Your implementation does not have this problem.

The multiprocessing pool is launched first and then CSRSelectors are created using freshly opened file instances inside the worker processes. This way, there will be no conflicts due to shared file state between processes. So this looks good! :)

Issue #361, opened 2022-05-31
saddle script error


Snehal1894:
While generating saddle plots using command line with --vrange option, the script gave an error- 'variable **hi** referenced before assigned'.
The script was missing the hi/lo intialization for vrange.

adding the following lines of code in api/saddle.py resolved the issue.

if vrange is not None:
        lo, hi = vrange
        binedges = np.linspace(lo, hi, n_bins + 1)

nvictus:
Are you still encountering this issue? The initialization seems to be there: https://github.com/open2c/cooltools/blob/master/cooltools/api/saddle.py#L230-L232.

Snehal1894:
I added the vrange condition in saddleplot function after line 655, then the error was resolved.
In line 653 qrange assignes hi, lo; but for vrange these variables are not getting intitalized I suppose.

gfudenberg:
closed in #361, with added test

nvictus:
> I added the vrange condition in saddleplot function after line 655

Thank you! It indeed was missing in that case. Resolved.

PR #375, opened 2022-08-11
Fix uninitialized binedges #361

nvictus:


gfudenberg:
I guess it just needs flake7 @nvictus ?

PR #378, opened 2022-08-19
Create virtual 4C tool

Phlya:
API and CLI for virtual 4C.
No support for expected (for now?).

### DIFF
--- a/cooltools/__init__.py
+++ b/cooltools/__init__.py
@@ -33,3 +33,4 @@ from .api.snipping import pileup
 from .api.directionality import directionality
 from .api.insulation import insulation
 from .api.dotfinder import dots
+from .api.virtual4c import virtual4c
--- a/cooltools/api/virtual4c.py
+++ b/cooltools/api/virtual4c.py
@@ -82,6 +82,11 @@ def virtual4c(
     v4C_table : pandas.DataFrame
         A table containing the interaction frequency of the viewpoint with the rest of
         the genome
+
+    Note
+    ----
+    Note: this is a new (experimental) function, the interface or output might change in
+    a future version.
     """
     if clr_weight_name not in [None, False]:
         # check if cooler is balanced
--- a/cooltools/api/virtual4c.py
+++ b/cooltools/api/virtual4c.py
@@ -29,7 +29,7 @@ def extract_profile(chrom, clr, clr_weight_name, viewpoint):
     pxls1[["end2"]] = viewpoint[2]
 
     pxls1 = (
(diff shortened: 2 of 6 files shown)


Phlya:
Example how to use. Notice that using a viewpoint larger than binsize averages all bins within it, and masks out interactions inside the viewpoint.
Screenshot 2022-08-19 at 11 02 28

gfudenberg (commented):
nice! 
Couple q's:
- would it make sense to add an option to smooth via https://github.com/open2c/cooltools/blob/51cc542365aedb8776d7fde1de1e8cd8d687b723/cooltools/lib/numutils.py#L994
- should this go through sandbox first?

Phlya:
Can try to see what smoothing does to this! But I'm not sure there really is a reason to smooth this, idk...

Don't mind putting into sandbox first!

Phlya:
For sandbox, is there a way to make the CLI accessible somehow?..

Phlya:
For CLI tests to pass I had to add an empty `return` int he end and assert that exit code==0, not ==1, any idea why? Otherwise the code was 2. In geneal, some of out tests assert code==0, and some==1. Why is that?

agalitsyna:
Exit codes seems like a serious issue, e.g. see here (exit_code==1, but not 0 for no reason): https://github.com/open2c/cooltools/blob/a86fbec9601459624568f7f03c6ec89d445ca416/tests/test_insulation.py#L23

**_This is dangerous for the tests_**, because if it's actual mistake it can also naturally have exit_code 1 and it will be never captured by our tests. 

One of the ways to debug it is to output stderr and stdout alongside with exit_code while creating tests: https://github.com/pallets/click/issues/737

Phlya:
I think in this case it's not an issue, since with an empty return statement at the end the code is 0. And I assume if there is an actual error it won't get to the return statement and will give an error-y code. So perhaps this should be a broader discussion in a separate issue? @agalitsyna

Phlya:
@gfudenberg added a note that it's new and experimental, do you think it's enough?

gfudenberg:
sounds good!

one question though: would these functions make more sense as part of cooler, if they don't modify/process in any way other than querying/balancing ?
cc @nvictus

Phlya:
It's also averaging bins within the viewpoint...

gfudenberg:
indeed!

Phlya:
@gfudenberg any more comments?

gfudenberg:
nope!

Phlya:
Ok to merge?

gfudenberg:
fine by me!

Issue #353, opened 2022-04-17
expected_cis: interplay between smoothing and aggregation options is surprising


sergpolly:
in particular - following combination of flags yields no extra columns in the output - instead it should probably `raise` ? or `warn` ...
```python
    smooth=False,
    aggregate_smoothed=True,
```

golobor:
sure, raising an Exception would be nice!

PR #380, opened 2022-08-25
update-behavior-of-expected_cis

gfudenberg:
previously 
```python
expected_cis(
    ...,
    smooth=False,
    aggregate_smoothed=True
)
```
would ignore `aggregate_smoothed=True`.

This update:
- allows these columns to be appended separately by removing a (probably unintended) indent. 
- propagates nans from balanced.avg to smoothed columns.

Addresses 
https://github.com/open2c/cooltools/issues/353

### DIFF
--- a/cooltools/api/expected.py
+++ b/cooltools/api/expected.py
@@ -1016,6 +1016,7 @@ def expected_cis(
         result[f"{key}.avg"] = result[f"{key}.sum"] / result[_NUM_VALID]
 
     # additional smoothing and aggregating options would add columns only, not replace them
+    smoothed_columns = []
     if smooth:
         result_smooth = expected_smoothing.agg_smooth_cvd(
             result,
@@ -1027,21 +1028,25 @@ def expected_cis(
             on=[_REGION1, _REGION2, _DIST],
             how="left",
         )
-        if aggregate_smoothed:
-            result_smooth_agg = expected_smoothing.agg_smooth_cvd(
-                result,
-                groupby=None,
-                sigma_log10=smooth_sigma,
-            ).rename(columns={"balanced.avg.smoothed": "balanced.avg.smoothed.agg"})
-            # add smoothed columns to the result
-            result = result.merge(
-                result_smooth_agg[["balanced.avg.smoothed.agg", _DIST]],
-                on=[
-                    _DIST,
-                ],
-                how="left",
-            )
-
+        smoothed_columns.append("balanced.avg.smoothed")
+    if aggregate_smoothed:
+        result_smooth_agg = expected_smoothing.agg_smooth_cvd(
+            result,
+            groupby=None,
+            sigma_log10=smooth_sigma,
+        ).rename(columns={"balanced.avg.smoothed": "balanced.avg.smoothed.agg"})
+        # add smoothed columns to the result
+        result = result.merge(
+            result_smooth_agg[["balanced.avg.smoothed.agg", _DIST]],
+            on=[
+                _DIST,
+            ],
+            how="left",
+        )
+        smoothed_columns.append("balanced.avg.smoothed.agg")
+    # propagate nan
+    for key in smoothed_columns:
+        result[key].iloc[ result['balanced.avg'].isna().values ] = np.nan
     return result
 
 


gfudenberg:
one thing we may or may not want to change is that the function currently only allows smoothing / aggregated_smoothing for the balanced.avg key (i.e. the DEFAULT_CVD columns, which are present in sandbox, are not exposed to the user in api.expected)
https://github.com/open2c/cooltools/blob/a86fbec9601459624568f7f03c6ec89d445ca416/cooltools/api/expected.py#L1026

thoughts? cc @golobor @sergpolly

sergpolly:
> one thing we may or may not want to change is that the function currently only allows smoothing / aggregated_smoothing for the balanced.avg key (i.e. the DEFAULT_CVD columns, which are present in sandbox, are not exposed to the user in api.expected)
> 
> https://github.com/open2c/cooltools/blob/a86fbec9601459624568f7f03c6ec89d445ca416/cooltools/api/expected.py#L1026
> 
> thoughts? cc @golobor @sergpolly

I think it is worth exposing in the API for the new release - i think hardcoded `balanced.avg` is just a historic artifact

On the PR itself - you basically allow for independent smoothing or smoothing-aggregated ... right? - one problem is, that the docstring is written such that `aggregate_smoothing` looks like a sub-flag of the `smooth-cis` itself . what's the most logical API ? from the user perspective ?

gfudenberg:
ah yeah, I didn't update docstring. I guess since they are separate columns, I could imagine a user wanting the aggregated smoothing and not the individual chrom smoothing, and vice versa?

sergpolly:
The way the flags are named right now - implies that aggregation is a "subflag" of smoothing - or at least - that's how I though about it.

I'm fine with the independent smoothing/aggregation API -   but we should rename flags then as well - IMHO

E.g.
```python
Smooth=True,
Smooth_and_aggregate=True  # not aggregate_smoothed
```
What do you think ?

gfudenberg:
yes, I agree the naming could be confusing as well. Maybe even worth trying to make clear these are returning extra columns?
e.g. 
 ```python
return_smoothed=True
return_aggregated_smoothed=True
```
?

**Comment by sergpolly** [comment]
Yes - that looks more explicit ! Nice ! These flags seem compatible with the independent behavior ...

Can go even further to say that aggregated is genome_wide_smoothed - which it is.

I think we should run this API by someone else to see the reaction,  what do you think?

**Comment by gfudenberg** [comment]
agreed!

**Comment by sergpolly** [comment]
thought about it, while working on `expected_full` (still in the sandbox) - here is an alternative API:

smooth and aggregate are independent but in a different sense:
```python
smooth = True | False,  # engages smoothing using agg_smooth_cvd function
aggregate = None(False) | "chrom" | "genome",  # enables aggregation on chromosome or genome-wide level regardless of smoothing
expected_column_name = "expected"  # column to store final expected in - to avoid balanced.avg.smooth balanced.avg.smooth.agg
```

in case of `expected_full` it would look like that:
```python
smooth_cis = True | False,  # engages smoothing using agg_smooth_cvd function
aggregate_cis = None(False) | "chrom" | "genome",  # enables aggregation on chromosome or genome-wide level regardless of smoothing
aggregate_trans = None(False) | "chrom" | "genome",  # enables aggregation on chromosome or genome-wide level
expected_column_name = "expected"  # column to store final expected in - cis and trans
```

sergpolly:
so, several use-cases of `expected_cis` would look like this:
```python
expected_cis(
    clr,
    view_df=hg38_arms,
    smooth=True,
    aggregate=False,
    expected_column_name="expected"
)
```
i.e. calculate your cis expected by arm, smooth it without any aggregation and store the result in column `expected`

radically different alternative would be:
```python
expected_cis(
    clr,
    view_df=hg38_arms,
    smooth=False,
    aggregate="genome",
    expected_column_name="expected"
)
```
i.e. calculate your cis expected by arm, and aggregate genome-wide without smoothing (average over all arms excluding inter-arm), store the result in a column `expected`

A typical use-case for `expected_full` would be like this:
```python
expected_full(
    clr,
    view_df=hg38_arms,
    smooth_cis=True,
    aggregate_cis=False,
    aggregate_trans="chrom",
    expected_column_name="expected"
)
```
i.e. calculate cis and trans expected by arms first, smooth cis part of the expected without any aggregation , but aggregate trans expected to the chromosome-level - store all of the results (cis and trans) in column named `expected`

gfudenberg:
for `expected_cis`, `aggregate=False` vs. `aggregate="genome"` could be a bit confusing... maybe just aggregate=True as an option which aggregates over all regions in view, and if you want to do something more complicated you have to deal with the expected dataFrame?

Issue #377, opened 2022-08-15
random-sample counts cis/total


efriman:
Hi,

Just a few points that I noticed. Maybe there are good reasons for this but just wanted to point it out.

1. The "cis" counts (generated by cooltools coverage --store) are doubled, while the "sum" values are not. It seems for cooltools random-sample you need to use the doubled value with --counts-cis and the non-doubled value with --counts. This is a bit confusing.

2. After using cooltools random-sample, the cooler needs to be rebalanced but the weight column persists. This leads for example to the error "ValueError: Please, specify ignore_diags and/or IC balance this cooler! Cannot access the value used in IC balancing." when running cooltools coverage on the sampled (but not rebalanced) matrix and cooler balance says it's already balanced. Maybe it would be good to remove the weight column upon sampling.

Elias

Phlya:
Indeed, 1 is a bug! Thanks for reporting.

2 is a good idea, we should simply remove the weights after sampling, I also think so.

PR #381, opened 2022-08-25
fix377

gfudenberg:
addresses https://github.com/open2c/cooltools/issues/377

- only passes ['chrom','start','end'] from previous cooler bintable
- divides cis-coverage by two, since the coverage function considers each side separately, vs. count which counts number of reads

### DIFF
--- a/cooltools/api/coverage.py
+++ b/cooltools/api/coverage.py
@@ -121,5 +121,5 @@ def coverage(
                 grp["bins"].create_dataset(
                     store_name, data=cov_arr, **h5opts, dtype=int
                 )
-            grp.attrs.create("cis", np.sum(covs[0]), dtype=int)
+            grp.attrs.create("cis", np.sum(covs[0]) // 2, dtype=int)
     return covs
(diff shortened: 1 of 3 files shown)


Phlya:
Thanks for addressing this! The non-doubling of coverage actually is an issue also in the `cooltools coverage`, and hence the stored "cis" value is also wrong by 2-fold...
https://github.com/open2c/cooltools/blob/a86fbec9601459624568f7f03c6ec89d445ca416/cooltools/api/coverage.py#L124

Phlya:
It's just to try and use the stored value if it's already calculated, and if not - calculate it.

gfudenberg:
ah, so if the //2 is added to api.coverage, then all should be good?

Phlya:
Yes!

gfudenberg:
ok! added new tests as well, that will also help clarify the behavior of `coverage()` on a very simple cooler.

Phlya:
Looks good to me, thank you!

PR #385, opened 2022-09-13
Added clr_weight_name argument to coverage to calculate balanced coverage

efriman:


gfudenberg (changes requested):
Hi @efriman ,
thank you for the commit! This looks like useful functionality. If you could update two things, we could merge it into the main branch:
- clr_weight_name default would be better as None rather than an empty string
- for the added weight name, add a corresponding check for is_cooler_balanced with that weight
- even though computing coverage for raw & non-raw at the same time is more efficient, the output is a bit less interpretable so it would be better to still output a 2xN bins matrix no matter how the function is called rather than sometimes outputting a 2xN and other times a 4xN.
- a test for the case where a string is provided for this weight with a very simple example where you can know the output, e.g. https://github.com/open2c/cooltools/blob/076a945b51b21744a3e0646f8452fa5aaa82cb91/tests/test_coverage.py#L33

efriman:
Hi @gfudenberg! I think I addressed everything. I still made it so the default output of store_names adds the name of the specified weight column to not cause confusion. Let me know if you agree.

cooltools/api/coverage.py
```diff
@@ -102,24 +125,40 @@ def coverage(
...
         )
-
+    if clr_weight_name is not None:
+        try:
+            _ = is_cooler_balanced(clr, clr_weight_name, raise_errors=True)
+        except Exception as e:
+            raise ValueError(
+                f"provided cooler is not balanced or {clr_weight_name} is missing"
+            ) from e
+
```
sergpolly, line 135: maybe here or in a separate `if clr_weight_name` clause we could add the `bias` and `na_mask` initialization, like so:
```python
if clr_weight_name is not None:
    bias = clr.bins()[clr_weight_name][:]
    bias_na_mask = np.isnan(bias)  # remember masked bins
    bias = np.nan_to_num(bias)  # replace NaNs with 0 in bias to enable summation
```

cooltools/api/coverage.py
```diff
@@ -102,24 +125,40 @@ def coverage(
...
+            _ = is_cooler_balanced(clr, clr_weight_name, raise_errors=True)
+        except Exception as e:
+            raise ValueError(
+                f"provided cooler is not balanced or {clr_weight_name} is missing"
+            ) from e
+            
     chunks = cooler.tools.split(clr, chunksize=chunksize, map=map, use_lock=use_lock)
 
     if ignore_diags:
         chunks = chunks.pipe(_zero_diags, n_diags=ignore_diags)
```
sergpolly, line 140: here we'd add the balancing transformation step to the `split-apply-combine` like so:
```python
if clr_weight_name is not None:
    pixel_weight_key = "balanced"
    chunks = chunks.pipe(_apply_balancing, bias=bias, balanced_column_name=pixel_weight_key')
else:
    pixel_weight_key = "count"

```

cooltools/api/coverage.py
```diff
@@ -102,24 +125,40 @@ def coverage(
...
 
     n_bins = clr.info["nbins"]
-    covs = chunks.pipe(_get_chunk_coverage).reduce(np.add, np.zeros((2, n_bins)))
-    covs = covs[0].astype(int), covs[1].astype(int)
+    get_chunk_coverage = partial(_get_chunk_coverage, clr_weight_name=clr_weight_name)
```
sergpolly, line 142: no need for `partial`, we could just go straight into:
```python
covs = chunks.pipe(_get_chunk_coverage, pixel_weight_key=pixel_weight_key).reduce( np.add, np.zeros(n_bins) )
```

cooltools/api/coverage.py
```diff
@@ -102,24 +125,40 @@ def coverage(
...
+    covs = chunks.pipe(get_chunk_coverage).reduce(np.add, np.zeros((2, n_bins)))
+    if clr_weight_name is None:
+        covs = covs[0].astype(int), covs[1].astype(int)
```
sergpolly, line 145: somewhere here we'd need to remeber to reapply `na_mask` kind of like:

```python
if clr_weight_name is not None:
    covs[:, bias_na_mask] = np.nan  # mask both columns of the balanced coverage
else:
    covs = covs.astype(int)
```
please note the vectorized form of numpy operations, IMHO they're more compact and more readable

cooltools/api/coverage.py
```diff
@@ -102,24 +125,40 @@ def coverage(
...
+    elif store_names == ["cis_raw_cov", "tot_raw_cov"]:
+        store_names = [s + str("_"+clr_weight_name) for s in store_names]
 
     if store:
         with clr.open("r+") as grp:
+            if clr_weight_name is None:
```
sergpolly, line 151: IMHO this is one of the rare occasions where a one-liner `if-else` works better:
```python
dtype = int if clr_weight_name is None else float
```

but that's just taste ...

cooltools/api/coverage.py
```diff
@@ -102,24 +125,40 @@ def coverage(
...
+                dtype = int
+            else:
+                dtype = float
             for store_name, cov_arr in zip(store_names, covs):
```
sergpolly, line 155: this would still work , even if one skips turning `covs` into tuple `covs = covs[0].astype(int), covs[1].astype(int)` - so `covs.astype(int)` is not a problem

cooltools/cli/coverage.py
```diff
@@ -80,15 +88,19 @@ def coverage(
...
             pool.close()
 
     coverage_table = clr.bins()[:][["chrom", "start", "end"]]
-    coverage_table["cis_raw_cov"] = cis_cov.astype(int)
-    coverage_table["tot_raw_cov"] = tot_cov.astype(int)
+    if clr_weight_name is None:
+        coverage_table["cis_raw_cov"] = cis_cov.astype(int)
+        coverage_table["tot_raw_cov"] = tot_cov.astype(int)
+    else:
+        coverage_table["cis_raw_cov"+str("_"+clr_weight_name)] = cis_cov.astype(float)
```
sergpolly, line 102: we have to figure out something with the naming, as `cis_raw_cov_weigtht` does not seem readable either ...

maybe names like these would be better:
```python
["cis_cov_raw", "tot_cov_raw"],  # raw names when using counts
["cis_cov_weight", "tot_cov_weight"],  # balanced names ...
```

so, maybe we could replace `store_names=["cis_raw_cov", "tot_raw_cov"],` argument in `coverage` function with something like `store_prefixes=["cis_cov", "tot_cov"]` and then add `_raw`/`_clr_weight_name` suffix to combine raw/balanced scenarios together  - what do you guys think ?

cooltools/api/coverage.py
```diff
@@ -55,6 +77,7 @@ def coverage(
     chunksize=int(1e7),
     map=map,
     use_lock=False,
+    clr_weight_name=None,
     store=False,
     store_names=["cis_raw_cov", "tot_raw_cov"],
```
sergpolly, line 82: consider replacing it with the `prefixes` parameter - see comments below

cooltools/cli/coverage.py
```diff
@@ -80,15 +88,19 @@ def coverage(
...
+        coverage_table["tot_raw_cov"+str("_"+clr_weight_name)] = tot_cov.astype(float)
```
sergpolly, line 103: we've been trying to switch to python f-strings in the cooltools , IMHO they're more readable :
```python
"tot_raw_cov"+str("_"+clr_weight_name)  # VS
f"tot_raw_cov_{clr_weight_name}"
```

sergpolly (changes requested):
hi, I've suggested some stylistic/implementation changes that are IMHO worth implementing
they are in the comments above
what do you guys think @efriman @gfudenberg ?

cooltools/cli/coverage.py
sergpolly: or maybe even simplier, like `store_prefix="cov"` and then just generate store name like so:
```python
if clr_weight_name is not None:
    store_names = [f"{store_prefix}_cis_{clr_weight_name}", f"{store_prefix}_tot_{clr_weight_name}" ]
else:
    store_names = [f"{store_prefix}_cis_raw", f"{store_prefix}_tot_raw" ]
```
or something like that

efriman:
Hey @sergpolly! Great suggestions. I think I've implemented them all. I hope I got it right, let me know. Tests run fine

tests/test_coverage.py
```diff
@@ -53,3 +53,30 @@ def test_coverage_symmetric_upper(request):
     assert (tot_cov == np.array([3, 1, 2])).all()
     assert clr.info["cis"] == 1
     assert clr.info["sum"] == 3
+    
+def test_coverage_weight(request):
```
sergpolly, line 57: maybe `test_balanced_coverage` ? it is usually called `weight` but people usually refer to "balancing" and balanced-data - it's more readable

cooltools/cli/coverage.py
```diff
@@ -49,6 +49,14 @@
...
 )
+@click.option(
+    "--clr_weight_name",
+    help="Name of the weight column. Specify if wanting to calculate"
+    " balanced coverage. In this case, cis counts will not be stored.",
+    type=str,
+    default=None,
+    show_default=False,
+)
 @click.option(
```
sergpolly, line 60: we might want to expose that prefix thing as a CLI argument  - maybe not - not sure ... just to provide flexible naming 
@gfudenberg @Phlya ?

cooltools/api/coverage.py
```diff
@@ -102,24 +114,48 @@ def coverage(
         raise ValueError(
             "Please, specify ignore_diags and/or IC balance this cooler! Cannot access the value used in IC balancing. "
         )
-
+    if clr_weight_name is not None:
```
sergpolly, line 117: somehow I missed it from the main review:
I wonder if we could streamline this clause in a simial manner as here https://github.com/open2c/cooltools/blob/2ba3c6ef6bb20b182b81f47ee03707298fa101fa/cooltools/api/expected.py#L1110

so it would be like so:
```python
    if clr_weight_name is None:
        # summing raw pixel counts
        pixel_weight_key = "count"
    elif is_cooler_balanced(clr, clr_weight_name):
        # initialize balancing bias and masking
        bias = clr.bins()[clr_weight_name][:]
        bias_na_mask = np.isnan(bias)  # remember masked bins
        bias = np.nan_to_num(bias)
        # summing balanced pixels
        pixel_weight_key = "balanced"
    else:
        raise ValueError(
            "cooler is not balanced, or"
            f"balancing weight {clr_weight_name} is not available in the cooler."
        )

```

cooltools/api/coverage.py
```diff
@@ -102,24 +114,48 @@ def coverage(
...
+        bias = clr.bins()[clr_weight_name][:]
+        bias_na_mask = np.isnan(bias)  # remember masked bins
+        bias = np.nan_to_num(bias)
+        
     chunks = cooler.tools.split(clr, chunksize=chunksize, map=map, use_lock=use_lock)
 
     if ignore_diags:
         chunks = chunks.pipe(_zero_diags, n_diags=ignore_diags)
 
+    if clr_weight_name is not None:
```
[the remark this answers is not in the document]
sergpolly: see a single-comment at the bottom for context - it disappeared from the review somehow

cooltools/api/coverage.py
```diff
@@ -82,8 +93,9 @@ def coverage(
         If None, equals the number of diagonals ignored during IC balancing.
     store : bool, optional
         If True, store the results in the file when finished. Default is False.
-    store_names : list, optional
-        Names of the columns of the bin table to save cis and total coverages.
+    prefixes : str, optional
```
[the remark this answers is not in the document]
gfudenberg: can we get rid of the `store` argument, and just have if `store_prefix=None` not save? or `save_prefixes=`?

tests/test_coverage.py
```diff
@@ -53,3 +53,30 @@ def test_coverage_symmetric_upper(request):
...
+
+    pixels = pd.DataFrame(
+        [[0, 1, 1], [0, 2, 2]], columns=["bin1_id", "bin2_id", "count"]
+    )
+
+    clr_file = op.join(request.fspath.dirname, "data/test_coverage.cool")
+    cooler.create_cooler(clr_file, bins, pixels)
+    clr = cooler.Cooler(clr_file)
+    cis_cov_weight, tot_cov_weight = cooltools.coverage(clr, ignore_diags=0, store=True, clr_weight_name="weight")
+    assert (cis_cov_weight == np.array([0.5, 0.5, 0])).all()
```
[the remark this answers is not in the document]
sergpolly: also for tests - should we add a bin or 2 with `NaN` in weights - just to make sure that bias masking mechanism is working ok  - in case we ever break it ...

cooltools/api/coverage.py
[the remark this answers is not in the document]
sergpolly: IMHO - `store`  is just simple and readable ...
and the `store_prefix` wouldn't be used often anyways, as long as the default prefix is "good"

If anything I would make `store` even more explicit - i.e. name it `store_in_cooler` or explain in the docstring that it is going to store results right in the cooler-file that was given as an input

cooltools/api/coverage.py
```diff
@@ -55,15 +64,17 @@ def coverage(
...
+    store_prefix="cov",
 ):
 
     """
     Calculate the sums of cis and genome-wide contacts (aka coverage aka marginals) for
     a sparse Hi-C contact map in Cooler HDF5 format.
     Note that the sum(tot_cov) from this function is two times the number of reads
-    contributing to the cooler, as each side contributes to the coverage.
+    contributing to the cooler, as each side contributes to the coverage 
+    (this only applies if clr_weight_name=None).
```
gfudenberg, line 77: this paranthetical is currently confusing to me.. would
Note that for raw coverage (i.e. `clr_weight_name=None`) then sum(tot_cov) calculated by this function is two times the number of reads  contributing to the cooler, as each side contributes to the coverage.

cooltools/api/coverage.py
gfudenberg: is the pixel weight always stored in 'balanced' or is it named after the specific weight clr_weight_name?

gfudenberg: also does it modify the input cooler? if so that would probably not be desired behavior?

cooltools/cli/coverage.py
```diff
@@ -49,6 +49,14 @@
     is_flag=True,
     default=False,
 )
+@click.option(
+    "--clr_weight_name",
+    help="Name of the weight column. Specify if wanting to calculate"
+    " balanced coverage. In this case, cis counts will not be stored.",
```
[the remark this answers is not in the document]
gfudenberg: How about this?
Name of the weight column. Specify to calculate coverage of a balanced cooler. If specified, then the cis and tot values for balanced coverage are stored rather than raw coverage.

cooltools/cli/coverage.py
gfudenberg: mmm, yeah seems complicated... 

changing store to store_in_cooler would definitely clarify the use of that argument vs. the output argument though for the CLI!

tests/test_coverage.py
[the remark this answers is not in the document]
gfudenberg: YES to adding bin with Nans
it would also be good to add test that covers the usage of the prefixes & storing

cooltools/api/coverage.py
sergpolly: No, here we define how to call balanced pixel values (kind of like in expected - transform function) and we use that name in the _apply_balancing function, regardless of how balancing weights are called in cooler

sergpolly: No no - balanced values don't get stored in the cooler - same as in expected summations - we apply balancing on the fly - via transform - use the values - and that's it - garbage collect takes over it I guess eventually

tests/test_coverage.py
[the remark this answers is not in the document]
sergpolly: The storing to cooler part, I believe is tested - in the test function above this one

cooltools/cli/coverage.py
```diff
@@ -80,15 +88,19 @@ def coverage(
...
-    coverage_table["tot_raw_cov"] = tot_cov.astype(int)
+    if clr_weight_name is None:
+        coverage_table["cov_cis_raw"] = cis_cov.astype(int)
+        coverage_table["cov_tot_raw"] = tot_cov.astype(int)
+    else:
+        coverage_table[f"cov_cis_{clr_weight_name}"] = cis_cov.astype(float)
+        coverage_table[f"cov_tot_{clr_weight_name}"] = tot_cov.astype(float)
 
     # output to file if specified:
     if output:
```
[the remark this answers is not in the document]
efriman: Good point. I changed how this is specified a bit to not have to repeat the assignment.  Let me know what you think

cooltools/api/coverage.py
efriman: Changed it. But isn't it also true for cis_cov, i.e.sum (cis_cov) is twice the number of cis reads? And if so should it say?

efriman:
I just noticed "use_lock" is not in the help for coverage.py. What does it do?

cooltools/api/coverage.py
efriman: Implemented your suggestion @sergpolly !

cooltools/cli/coverage.py
efriman: Yeah I think it could be nice to be able to specify in CLI. Let me know if others agree and I can add it.

cooltools/cli/coverage.py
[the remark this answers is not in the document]
efriman: Okay I've changed the help to be more explicit about this for store=True. Let me know

tests/test_coverage.py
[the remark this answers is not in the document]
efriman: Yes good idea with allclose and nans. Implemented this, let me know!

cooltools/api/coverage.py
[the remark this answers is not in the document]
sergpolly: here `else` part isn't needed anymore , and `pixel_weight_key` is already assigned - so you could just say:
```python
if clr_weight_name is not None:
    chunks = chunks.pipe(_apply_balancing, bias=bias, balanced_column_name=pixel_weight_key)
```

cooltools/cli/coverage.py
```diff
@@ -30,8 +30,9 @@
 )
 @click.option(
     "--store",
-    help="Append columns with coverage (cis_raw_cov, tot_raw_cov) "
-    " to the cooler bin table.",
+    help="Append columns with coverage (cis_raw_cov, tot_raw_cov)"
```
sergpolly, line 33: column names are different now - let's update them here as well

cooltools/cli/coverage.py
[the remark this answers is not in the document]
sergpolly: Thank you ! Look good to me

sergpolly (changes requested):
I think we're almost there ! There as one `if`-clause to change and the API/CLI decision ...

Maybe we could accept as is - and decide on the API/CLI later - maybe this would be most productive @gfudenberg - I have no strong opinion on that

sergpolly:
> I just noticed "use_lock" is not in the help for coverage. What does it do?

I believe this is related to hdf5, which cooler is based on - like, can multiple processes access the same 'file" or not, but i'm not 100% sure

cooltools/api/coverage.py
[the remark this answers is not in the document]
efriman: Yes I just realized this after my first commit :) Thanks

efriman:
How about removing use_lock from the function arguments and setting use_lock=False in the .split?

`chunks = cooler.tools.split(clr, chunksize=chunksize, map=map, use_lock=False)`

cooltools/api/coverage.py
gfudenberg: Counting each side individually happens regardless of clr_weight_name. It's just that when clr_weight_name=None then it is calculating on raw reads, and this two times the number of reads holds exactly. Hence the previous statement "(this only applies if clr_weight_name=None)" is worded confusingly.

p.s. I'm not seeing this suggested docstring change when I view the file

efriman: This is so weird... It's in my cloned repository

efriman: Also I just meant that the cis counts from raw coverage will be twice the total number of cis counts in the cooler. Not just total

efriman: Any idea why this is not showing up in the PR? It's here: https://github.com/efriman/cooltools/blob/master/cooltools/api/coverage.py#L78

Not sure what to do

sergpolly:
tests were failing because of some other function - so we merged.
Thank you @efriman !

efriman:
Great! Thanks a lot for the input and for all the great work you do

PR #400, opened 2022-11-01
Fixed bug in io.read_expected_from_file when having mix of str and int as region names

efriman:


gfudenberg:
hi @efriman -- thanks for the PR! 
is there any chance you can add this to a fresh commit? It looks like your commits also reflect updates to coverage on your branch?
if you could add a new ./tests/test_io.py that covers this case that would be appreciated as well.

Issue #384, opened 2022-09-05
Expected with integer chromosomes


efriman:
Hi,

I noticed if I have cooler with chromosomes named without "chr", i.e. 1,2,3 etc., I can generate an expected file that looks right but when I run cooltools.lib.io.read_expected_from_file it throws an error. If I do it in the API it works if I convert the region1 and region2 columns to str.

Elias

gfudenberg:
@sergpolly should `read_expected_from_file` just used the dtype= kwarg for read_table before verifying? https://github.com/open2c/cooltools/blob/880a1487c66a404e88437fd697caf4cd91dc0004/cooltools/lib/io.py#L315

gfudenberg:
closed with #404

Issue #334, opened 2022-03-01
wow, thanks for this update @sergpolly, this looks great !!!


sergpolly:
@gfudenberg :
> update the `read_expected_from_file()` docstring to point to the schema https://github.com/open2c/cooltools/blob/2c17e395a4e252cf78ee9127d1de1d1b49e0e475/cooltools/lib/io.py#L290, and copy an up-to-date example from the issue to schemas

sergpolly:
this likely affect more tools than just dotfinder, so I made a separate issue

gfudenberg:
was the proposal to copy an example to lib/schemas or lib/io.read_expected_from_file? @sergpolly 
or just add header lines to read_expected_from_file and then the full example to schemas?
(and delete https://github.com/open2c/cooltools/blob/48bec4008d765b9e0ee0a72782a04d4bca73e930/cooltools/lib/io.py#L290)

In either case, can the example be: 
```python
region	diag	n_valid	count.sum	balanced.sum	count.avg	balanced.avg
chr2_p	0	8771	nan	nan	nan	nan
chr2_p	1	8753	nan	nan	nan	nan
chr2_p	2	8745	2659034.0	406.01308769383326	304.063350485992	0.046428026037030674
chr2_p	3	8741	1564890.0	237.58527079797543	179.0287152499714	0.027180559523850293
```

sergpolly:
wow - it's been awhile ...

I think it was just about adding an example to the function itself lib/io.read_expected_from_file
I think it is for user's understanding and users are unlikely to look in schemas 

So yes, we could just add this example - just need to make sure it is formatted nicely - numpy-style

PR #404, opened 2022-12-01
read_expected_from_file set dtypes

efriman:
Addresses issues:
https://github.com/open2c/cooltools/issues/384
https://github.com/open2c/cooltools/issues/334

read_expected_from_file takes dtypes from schemas which avoids certain errors when not setting dtypes. Added test to throw error if dtype is incompatible (str->int).

Any comments welcome

### DIFF
--- a/cooltools/lib/io.py
+++ b/cooltools/lib/io.py
@@ -289,6 +289,8 @@ def read_expected_from_file(
     Expected must conform v1.0 format
     https://github.com/open2c/cooltools/issues/217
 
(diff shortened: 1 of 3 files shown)


gfudenberg (changes requested):
looks good, thank you!! 
two more small things: 
- can you move the new test to a separate test_io.py file? 
- add a test that correctly passes when chromosomes don't have 'chr' (i.e. the issue you had previously if I recall)?

efriman:
@gfudenberg I implenented your suggestion, with a mix of str and int for region names which was the bug I was experiencing

gfudenberg (approved):
awesome, thank you!

Issue #406, opened 2022-12-06
pileup returns nans as zeros


gfudenberg:
Is this the intended behavior? @Phlya @agalitsyna 
https://github.com/open2c/cooltools/blob/48bec4008d765b9e0ee0a72782a04d4bca73e930/cooltools/api/snipping.py#L988

Here's a minimal example of what happens when`np.nan` entries are input to that line:
```python
stack = np.array([[[np.nan]]])
stack = np.array([[[np.nan]]])
stack.shape
(1, 1, 1)

np.nansum([stack, np.transpose(stack, axes=(1, 0, 2))], axis=0)
array([[[0.]]])
```

Phlya:
Mhhh good point... I think this is not intended behaviour, this is there to fill in the bottom triangle of the on-diagonal pileups, since otherwise they would be empty and people would be surprised. I guess the positions of NaNs in the upper triangle need to be recorded, and then equivalently transposed, and set again in the new array... Something like that.

golobor:
A potentially simpler solution:
https://github.com/open2c/cooltools/compare/master...symmetrize-pileups

see an example here:
https://gist.github.com/golobor/cdf86509191515b6e6ae3e4cd6a2f086

Phlya:
Cool, never heard of this function. Need to test, but looks promising!

PR #418, opened 2023-02-09
Record and restore nans when transposing the stack in pileups

Phlya:
Fix https://github.com/open2c/cooltools/issues/406

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,5 @@ tmp.npz
 .gitignore
 tmp.hdf5
 cooltools/sandbox/test.mcool
+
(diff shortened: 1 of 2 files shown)


Phlya:
Add np.triu before transposing to ensure the main diagonal is not summed up with itself in case it's not masked.

Phlya:
Maybe we can just to np.sum instead of np.nansum with nan-flipping?

Phlya:
Investigate why the stack is organised with indices along the third axis, not first?

Should the transposing/summing be a separate small function?

Issue #373, opened 2022-08-03
Zeros in expected table: causing error for dot finder and other tools


agalitsyna:
**_Precedent:_** Expected table with zeros will cause NaNs in the dotfinder tables, which will cause an error with poor description and hard to debug. 

**_Proposed generalized solution:_** 
1. Add checks for presence of zeros in the expected table when parsing it in the tools; throw a warning
2. Add handling zeros in the expected table when it can allow the tool work on general

**_Potential inspiration for solving:_** 
Separating truly missing data and known missing data in saddles/compartments when aligning tracks with coolers: https://github.com/open2c/cooltools/pull/360

agalitsyna:
Related report with the same issue in dot caller: https://github.com/open2c/cooltools/issues/364

sergpolly:
@gibcus in the lab discovered a related issue independently - i'll try to get his data to reproduce the error and than fix it. Preliminary this might be caused by the tiny arms of acrocentric chromosomes that have very little to no data associated with them

sergpolly:
addressed in dots with the latest commit https://github.com/open2c/cooltools/commit/d4518a1ec7e7cf25cf8a8ecc98d0a627cc2be90a

gibcus:
Getting a different error now:
```
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/gibcusj/miniconda3/envs/open2c-env/lib/python3.8/site-packages/pandas/core/indexing.py", line 1587, in _get_list_axis
    return self.obj._take_with_is_copy(key, axis=axis)
  File "/home/gibcusj/miniconda3/envs/open2c-env/lib/python3.8/site-packages/pandas/core/series.py", line 945, in _take_with_is_copy
    return self.take(indices=indices, axis=axis)
  File "/home/gibcusj/miniconda3/envs/open2c-env/lib/python3.8/site-packages/pandas/core/series.py", line 930, in take
    new_index = self.index.take(indices)
  File "/home/gibcusj/miniconda3/envs/open2c-env/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 1189, in take
    taken = values.take(
  File "/home/gibcusj/miniconda3/envs/open2c-env/lib/python3.8/site-packages/pandas/core/arrays/interval.py", line 1081, in take
    left_take = take(
[...]

Issue #352, opened 2022-04-14
Error when using the max-nans-tolerated option


jclqrs:
I'm trying to run the CLI cooltools dots function and it gives an error when I set the max-nans-tolerated option to 7.  The dots function runs properly when I get rid of the max-nans-tolerated option but keep everything else the same. This happens with cool files that I generated using the distiller pipeline.

Also, I wondering if the first 2 diagonals are considered as nan values? 

Here is the output:

>cooltools dots -o Y7_0h_midg1_rep0to3_10k_loops_nan7_FDR05 -p 16 --fdr 0.05 --view /home/lamj2/mi_tools/mm9_chr1toX.bed --max-nans-tolerated 7 ./Y7_0h_midg1_rep0to3.mcool::/resolutions/10000 Y7_0h_midg1_rep0to3_10k_expected-cis.tsv
INFO:root:Using recommended donut-based kernels with w=5, p=2 for binsize=10000
INFO:root: matrix 19720X19720 to be split into 1089 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 18175X18175 to be split into 961 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 15960X15960 to be split into 729 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 15564X15564 to be split into 676 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 15254X15254 to be split into 676 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 14952X14952 to be split into 625 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 15253X15253 to be split into 676 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 13174X13174 to be split into 484 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 12408X12408 to be split into 441 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
INFO:root: matrix 13000X13000 to be split into 484 tiles of 600X600.
INFO:root: tiles are padded (width=5) to enable convolution near the edges
[...]

PR #432, opened 2023-04-11
Dotsfix to prevent more IndexErrors

sergpolly:
small fixes in dotfinder to prevent IndexErrors in `extract_scored_pixels`: prevent `NaN`- valued pixel scores, consider "non-monotonous" histogram


sergpolly:
this should close finally #373 and potentially #352

Issue #421, opened 2023-02-14
cooltools expected with regions fails if regions do not have names


agalitsyna:
My regions table has three columns (chrom-start-end with no names).
expected-cis CLI fails with:

<img width="1024" alt="image" src="https://user-images.githubusercontent.com/8100372/218880857-594bae9a-7e7f-422e-8d9d-92f168e495cc.png">

sergpolly:
For a second thought we agreed on this - but upon re-reading the docstring realized that it isn't the case ...
So - you're right this should not fail like that !

I suspect the problem is with the the `bioframe.read_table` over [here](https://github.com/open2c/cooltools/blob/67bc94d41b988d4442f483da80c6274e91bda164/cooltools/lib/io.py#L372) - i.e. we try to read a `bed4` from a 3-column file and it should FAIL, but instead it is giving us a 4-column dataframe , where names are popluated with NaNs

need more insight into bioframe ...
ok - looks like we need to engage ` schema_is_strict=True` for this to work reliably !

PR #437, opened 2023-04-19
fix 421 - allow view without names

sergpolly:
engaged `schema_is_strict=True` for reading `bed4` or `bed3` viewframe from file addressing #421

I checked and it ooks like this is the only place where we use `bioframe.read_table` so this should be ok from generality perspective


sergpolly:
alright - i think i'll merge it as it is small, self-explanatory and mostly good

Issue #435, opened 2023-04-18
bioframe 0.4.0 "select" incompatibilities


sergpolly:
tests are failing right now because of this #433 #434 and also because of some `bioframe.select` incompatibilities in 0.4.0
in a nutshell:
with bioframe 0.3.1 , cooler from tests `tests/data/CN.mm9.1000kb.cool` and track is just an EV from said cooler - same 2666 bins:
``` python
In [15]: clr.matrix().fetch(("chr1",100000000,150000000) ).shape
Out[15]: (50, 50)

In [16]: bioframe.select(track, ("chr1",100000000,150000000)).shape
Out[16]: (50, 4)
```
and with bioframe 0.4.0:
```python
In [4]: bioframe.select(track, ("chr1",100000000,150000000)).shape
Out[4]: (51, 4)
```

sample output of the bioframe.select here:
![image](https://user-images.githubusercontent.com/6790270/232927465-f9bf3ef6-8aaa-479c-979a-3d6cabb43f07.png)

that *99th* bins is not included with the older bioframe

gfudenberg:
thanks for tracking this down!
Seems like this was addressed as Bug fix: treat query interval as closed on left
https://github.com/open2c/bioframe/commit/c22b6c20aeade98e82ca2b349a4b3d3651b22435
Which seems generally like the desired behavior... but I agree it is a bit weird for tracks.
Thoughts @nvictus  ?

nvictus:
I think the "bug fix" was a mistake. Assuming (1) the query interval is to be interpreted as half-open, closed on the left and (2) the target intervals are also half-open, closed on the left, and (3) we want this operation to return any target interval that **overlaps** the query interval. Then, counterintuitively, the search query should look like:

```
mask = (query_chrom == df.chrom) & (query_start < df.end) & (query_end > df.start)
```

Typical case with bookended target intervals:
```
---
                  [QQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQQ)
|      A       |        B        |      C      |       D        |         E         |
---
desired result: B, C, D
```

Edge-cases:
```
1. Query start position matches start of an interval
---
               [QQQQQQQQQQQQQQQQQQQQQQQQQ)
|      A       |        B        |      C      |       D        |
---
desired result: B, C (but not A)

2. Query end position matches end of an interval
---
       [QQQQQQQQQQQQQQQQQQQQQQQQQ)
|      A       |        B        |      C      |       D        |
---
[...]

PR #438, opened 2023-04-23
Pin bioframe >=0.4.1

nvictus:
Solves #435


sergpolly:
Awesome, thanks! Sorry I was away from 💻

PR #442, opened 2023-05-16
Update cross_score.py

Phlya:
Allow arbitrary clr-weight-name and mask first diagonals (default 2)


Phlya:
Would be good to add the code to default ignore_diags to the diags ignored in balancing

golobor (approved):
looks good to me!

PR #448, opened 2023-06-09
New tool: rearrange cooler

Phlya:
Both API and (first draft of) CLI to rearrange a cooler following a genomic view (almost like liftover).

Need to move the example notebook from sandbox to the main examples at some point...


cooltools/api/rearrange.py
```diff
@@ -178,7 +181,7 @@ def reorder_cooler(
     if new_chrom_col not in view_df.columns:
```
golobor, line 181: - throw a Warning if new_chrom_col name is not provided, but the default value is taken. 
- not sure about randint, but may be ok (this is an unlikely scenario anyway)

cooltools/api/rearrange.py
```diff
@@ -92,7 +95,7 @@ def _flip_bins(regdf):
     return regdf
 
 
-def _reorder_bins(
+def _rearrange_bins(
```
golobor, line 98: this is useful/general enough to drop the underscore and add a docstring

golobor: would this function be shorter if one were to use a lookup table to convert old bin_ids-> new bin_ids? Such a lookup table can be easily generated in rearrange_bins

Phlya:
Thank you for the idea @golobor, mapping bin IDs from old to new instead of complicated arithmetic is way simpler! I don't think there is a big difference in performance.

cooltools/api/rearrange.py
Phlya: Added the warning, but actually this can only happen when the user explicitly provides new_chrom_col=None or orientation_col=None, since the arguments have default values.

tests/test_reorder_cooler.py
agalitsyna, line 1: If we keep "reorder" in the sandbox, we may want to keep the tests for the sandbox separately.

agalitsyna (changes requested):
My tests are running perfectly, and the tool looks ready to become part of the main cooltools. If we don't do that for now, we should somehow separate tests for the sandbox from the main ones.

tests/test_reorder_cooler.py
Phlya: oh do you mean the name of the file with the tests? I just forgot to change it... Or what do you mean?

agalitsyna: If we don't include it as a main tool, maybe running this test conditionally on request is better? E.g. if we introduce some big change that affects everything, we don't need to debug the sandbox in the first place.

Phlya: But it's moved to the main tools in this PR, I don't see any reason not to do it. Why not just run this test every time?

agalitsyna: Ah, if you think it's ready for the main tools, then no change is needed, of course.

Phlya: I hope it's ready or nearly ready! I think we could move the example notebook somewhere more discoverable, otherwise it seems to work ax expected.
Maybe one more thing is to check the case when view regions are not bin-aligned...

Phlya:
@nvictus I now specify mergebuf here because I had issues with truncated files with the warning about it, as I mentioned at some point. Do you think it's OK like this? Or should it be an argument?

Phlya:
@agalitsyna @golobor it's been a long time, but I think this can be merged. The example notebook is not very discoverable at the moment, but I think it should be expanded into a more general "how to manipulate coolers" notebook and be part of open2c_examples, and I don't think that's part of this PR... After merging we can simply move it there as is for now though.

PR #451, opened 2023-06-30
nopython=True due to deprecation warning

efriman:
See https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-object-mode-fall-back-behaviour-when-using-jit


Phlya:
Looks good, I'll merge - thank you!

PR #464, opened 2023-10-18
Add pixel_filter_util.py and pool_decorator function

Yaoyx:
- add pixel_filter_util.py to sandbox. The script includes functions for generating a filtered cool file based on cis-total ratio threshold
- add pool_decorator function to lib/common.py. A decorator function that enables multiprocessing for a given function.


cooltools/sandbox/pixel_filter_util.py
```diff
@@ -0,0 +1,214 @@
...
+    chunk : numpy.ndarray
+        A chunk of pixels of size `chunksize`.
+    """
+    selector = clr.pixels()
+    for lo, hi in cooler.util.partition(0, len(selector), chunksize):
+        chunk = selector[lo:hi]
+        yield chunk
+
+
+def pool_decorator(func):
```
gfudenberg, line 135: Can probably delete this one & import from lib?

gfudenberg:
Would be great to add ipynb with example usage of this `cis_total_ratio_filter`. Perhaps this can be put in a subdirectory of sandbox/cooler_filters

gfudenberg:
@agalitsyna would you mind taking a look, since you mentioned this functionality could be useful for you?

cooltools/sandbox/pixel_filter_util.py
Yaoyx: updated

cooltools/lib/common.py
```diff
@@ -506,3 +508,39 @@ def align_track_with_cooler(
...
+    -------
+    A wrapper function that enables multiprocessing for the given function.
+    """
+
+    def wrapper(*args, **kwargs):
+        pool = None
+        if "nproc" in kwargs.keys():
+            if kwargs["nproc"] > 1:
+                pool = Pool(kwargs["nproc"])
+                mymap = pool.map
```
nvictus, line 532: `mp.Pool.map` is not always what you want to use in the decorated function. It is **eager** and will [materialize all outputs in a list](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.map).

This is problematic if the function you are mapping is not reductive (e.g. when making a pixel chunk iterator, all chunks will be materialized in memory). Instead you want to use a lazier map implementation like `imap` or in some cases `imap_unordered`.

cooltools/sandbox/cooler_filters/pixel_filter_util.py
```diff
@@ -0,0 +1,176 @@
...
+logger = logging.getLogger("data_util")
+logger.propagate = False  # Disable propagation to the root logger
+
+ch = logging.StreamHandler()
+ch.setLevel(logging.INFO)
+ch.setFormatter(formatter)
+logger.addHandler(ch)
+
+@curry
+def cis_total_ratio_filter(clr, thres=0.5):
```
gfudenberg, line 20: I would go with just `threshold=0.5` instead of thres, because it may be easier to remember (e.g. I would forget if I should have thres or thresh)

cooltools/sandbox/cooler_filters/pixel_filter_util.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+    logger.debug("Start to create cooler file...")
+    bin_table = clr.bins()[:]
+    good_bins_index = np.array(range(clr.bins().shape[0]))[bin_mask]
+    pixels_filter = functools.partial(_pixel_filter, good_bins_index=good_bins_index)
+
+    cooler.create_cooler(
+        output_uri,
+        bins=bin_table,
+        pixels=map(pixels_filter, pixel_iter_chunks(clr, chunksize)),
```
nvictus, line 171: You probably want either builtin `map`, which is lazy and returns an iterator, or `mp.Pool.imap` + `ordered=True`, or potentially `mp.Pool.imap_unordered` + `ordered=False` (this does a 2-pass creation algorithm).

cooltools/lib/common.py
gfudenberg: option for a test for this function could be returning what map is being used by it...

gfudenberg: seems like cooler indeed used pool.imap for tabix loader: https://github.com/open2c/cooler/blob/857601826fb3aa8ae4e4d3cc64afa61dec18c87e/src/cooler/cli/cload.py#L222

gfudenberg: pool.map was used for balance b/c the length bintable is reasonable to hold in memory, as are various function outputs from cooltools

cooltools/sandbox/cooler_filters/pixel_filter_util.py
```diff
@@ -0,0 +1,176 @@
...
+    Returns
+    -------
+    None
+    """
+    if len(bin_mask) != clr.bins().shape[0]:
+        raise ValueError(
+            "bin_mask should have the same length as bin table in cool file"
+        )
+    logger.debug("Start to create cooler file...")
+    bin_table = clr.bins()[:]
```
gfudenberg, line 164: we should drop fold weights from the bin_table to minimize confusion!

Yaoyx:
Updates:
- Change pool.map to pool.imap in pool_decorator
- rename <thres> to <threshold> in cis_total_ratio_filter
- remove weight column from the bin table in create_filtered_cooler

agalitsyna (approved):

Issue #472, opened 2024-01-05
Adaptive coarsegraining doesn't work with NumPy 1.24 and later


gschuette:
Hi, 

The adaptive_coarsegrain function doesn't work when using NumPy 1.24 and later because line [1318](https://github.com/open2c/cooltools/blob/8af2b1087f7302d282dc77a82047ac0a3a8339c1/cooltools/lib/numutils.py#L1318C1-L1319C1) in cooltools/cooltools/lib/numutils.py uses `np.int`, which was deprecated in NumPy 1.20 and fully removed in NumPy 1.24 (see [here](https://numpy.org/devdocs/release/1.20.0-notes.html)). It should be replaced by `int` or `np.int_`. 

Best, 
Greg

nvictus:
Thanks, Greg. If you like, you're welcome to initiate the PR yourself for a little GitHub karma. I would change it to `int`.

gschuette:
Thanks for the suggestion! I just submitted the pull request now.

PR #473, opened 2024-01-05
Replaced np.int with int in adaptive_coarsegrain

gschuette:
The adaptive_coarsegrain function doesn't work when using NumPy 1.24 and later because line [1318](https://github.com/open2c/cooltools/blob/8af2b1087f7302d282dc77a82047ac0a3a8339c1/cooltools/lib/numutils.py#L1318C1-L1319C1) in cooltools/cooltools/lib/numutils.py uses `np.int`, which was deprecated in NumPy 1.20 and fully removed in NumPy 1.24 (see [here](https://numpy.org/devdocs/release/1.20.0-notes.html)). It should be replaced by `int` or `np.int_`. 

Per the suggestion from @nvictus in issue [472](https://github.com/open2c/cooltools/issues/472), I opted for the `int` option.


nvictus:
Thank you @gschuette !

PR #477, opened 2024-01-16
enforce multiprocess-dill

sergpolly:
[`multiprocess`](https://pypi.org/project/multiprocess/) has been used in most of the cooltools so far (and in cooler itself) - imho we should continue using it throughout
examples where `multiprocessing` fail:
 - dots simply not work with it
 - multiprocessing prevents nested multiprocessiong https://github.com/open2c/open2c_examples/issues/29
 - more (?)


sergpolly:
closing in favor of #489 -> that'll take care of multiprocessing/pickle -> multiprocess/dill transition

Issue #431, opened 2023-04-06
multiprocessing consistency across cooltools


gfudenberg:
many user-facing functions take nproc as an argument. 
e.g. 
https://github.com/open2c/cooltools/blob/c85d647239934b290c844d4db46782aff71db04d/cooltools/api/snipping.py#L824

`cooltools.sample()` does not
https://github.com/open2c/cooltools/blob/c85d647239934b290c844d4db46782aff71db04d/cooltools/api/sample.py#L54
 (however, multiprocessing is offered as an nproc argument in=cooltools.cli.random-sample does)

Any suggestion on preferred behavior?

Phlya:
Can we make this into a decorator to ensure the argument is consistent across all tools and stays like that in the future?

gfudenberg:
note 10/23/2023: some of the cootlools functions using pool.map should be changed to pool.imap! 

see @nvictus  comment https://github.com/open2c/cooltools/pull/464#discussion_r1369165950

gfudenberg:
in the process of addressing this issue, @Yaoyx ran into a design choice:
(a) keep multiprocessing & factor out internal lambda functions (which pickle fails on)
(b) move to multiprocess (which uses dill instead of pickle) and otherwise keep code the same

any recommendations @golobor @nvictus @agalitsyna @sergpolly ?
an example of an internal lambda function that would need to be moved is here, inside of dots `scoring_and_histogramming()`: 

``` job = lambda tile: to_hist(to_score(tile))```
https://github.com/open2c/cooltools/blob/81b9045c58a59b374d0de5137f6c812f0a34f3ab/cooltools/api/dotfinder.py#L1307

sergpolly:
imho multiprocess/dill is **much** more preferable, because multiprocessing/pickle prevents one more thing: https://github.com/open2c/open2c_examples/issues/29

also I've been trying to hint at that here https://github.com/open2c/cooltools/pull/477

nvictus:
Multiprocessing will [eventually](https://github.com/python/cpython/issues/72240) let you register custom serializers, but until that happens I would stick with multiprocess+dill or another 3rd party pool manager like [loky](https://github.com/joblib/loky) or [mpire](https://github.com/sybrenjansen/mpire).

nvictus:
We might also want to explore changing the default context (which is `"fork"` on linux) to:

```python
from multiprocess import get_context

with get_context("spawn").Pool() as pool:
    ....
````
[...]

PR #489, opened 2024-01-22
Add pool decorator to functions

Yaoyx:


cooltools/api/dotfinder.py
```diff
@@ -1402,21 +1401,18 @@ def scoring_and_extraction_step(
...
-        filtered_pix_chunks = map_(job, tiles, **map_kwargs)
-        significant_pixels = pd.concat(filtered_pix_chunks, ignore_index=True)
-    finally:
-        if nproc > 1:
-            pool.close()
+    
+    map_ = map
+    
+    # consider using
+    # https://github.com/mirnylab/cooler/blob/9e72ee202b0ac6f9d93fd2444d6f94c524962769/cooler/tools.py#L59
```
gfudenberg, line 1412: is this deprecated @nvictus

cooltools/api/snipping.py
```diff
@@ -44,6 +44,7 @@
     is_valid_expected,
 )
 from ..lib.common import assign_view_auto, make_cooler_view
+from ..lib.common import pool_decorator
```
gfudenberg, line 47: this could be combined into previous line @Yaoyx

cooltools/api/insulation.py
```diff
@@ -17,6 +17,8 @@
 
 from ..lib.checks import is_compatible_viewframe, is_cooler_balanced
 from ..lib.common import make_cooler_view
+from ..lib.common import pool_decorator
```
gfudenberg, line 20: can be condensed: from ..lib.common import make_cooler_view, pool_decorator

cooltools/lib/common.py
```diff
@@ -532,15 +532,14 @@ def wrapper(*args, **kwargs):
...
+            return result 
         else:
-            raise TypeError(
-                "nproc must be specified as a keyword argument (e.g. nproc=1, nproc=5...)"
-            )
-
-        if pool != None:
-            pool.close()
-
+            warnings.warn("nproc is not specified, single thread is used then (e.g. nproc=1, nproc=5...)")
```
gfudenberg, line 542: this warning may be rephrased

cooltools/cli/coverage.py
```diff
@@ -67,8 +70,9 @@
     default=1,
     type=int,
 )
+@pool_decorator
```
gfudenberg, line 73: I wonder if the pool_decorator should instead be applied to the api version of this function? cc @Phlya

cooltools/cli/sample.py
```diff
@@ -5,6 +5,7 @@
 
 from . import cli
 from .. import api
+from ..lib.common import pool_decorator
```
gfudenberg, line 8: same question-- maybe more clear if the pool decorator is always in API rather than CLI version of the user-facing function?

gfudenberg:
the decorator does nicely streamline the code! any ideas on a potential tests to add for the `pool_decorator` function @Yaoyx ?

cooltools/api/snipping.py
Yaoyx: combined

tests/test_virtual4c.py
```diff
@@ -40,3 +40,16 @@ def test_virtual4c_cli_nobalance(request, tmpdir):
...
+
+
+def test_pooled_virtual4c(request):
+    clr_path = op.join(request.fspath.dirname, "data/CN.mm9.1000kb.cool")
+    clr = cooler.Cooler(clr_path)
+    viewpoint = "chr1:30000000-40000000"
+
+    v4c = virtual4c.virtual4c(clr, viewpoint)
+    pooled2_v4c = virtual4c.virtual4c(clr, viewpoint, nproc=2)
+    pooled3_v4c = virtual4c.virtual4c(clr, viewpoint, nproc=3)
```
gfudenberg, line 52: probably only need one of these if they cover the same argment?

tests/test_snipping.py
```diff
@@ -487,3 +487,141 @@ def test_cooler_snipper_with_view(request):
...
+    # Example on-diagonal features, not valid bedframes (start>end):
+    windows = pd.DataFrame(
+        {
+            "chrom": ["chr1", "chr1"],
+            "start": [107_000_000, 108_000_000],
+            "end": [102_000_000, 113_000_000],
+        }
+    )
+    with pytest.raises(ValueError):
+        stack = cooltools.api.snipping.pileup(clr, windows, view_df, exp, flank=None)
```
gfudenberg, line 591: probably only need one of these 3 tests-- also they are under the same with statement

tests/test_snipping.py
```diff
@@ -487,3 +487,141 @@ def test_cooler_snipper_with_view(request):
         matrix, "foo", "foo", (110_000_000, 120_000_000, 110_000_000, 120_000_000)
     )
     assert snippet.shape is not None
+
+def test_pooled_pileup(request):
```
gfudenberg, line 491: I wonder if we could just add 
```
stack_pooled2 = cooltools.api.snipping.pileup(clr, windows, view_df=None, flank=None, nproc=2)
assert np.array_equal(stack, stack_pooled2, equal_nan=True)
```

to the earlier test, rather than needing to recalculate all of this in the test_pooled_pileup

tests/test_insulation.py
```diff
@@ -176,3 +176,43 @@ def test_insulation_sparse_vs_dense(request):
...
+    insulation_pooled3 = calculate_insulation_score(
+        clr, 10_000_000, min_dist_bad_bin=0, view_df=region, nproc=3
+    )
+    assert insulation.equals(insulation_pooled2)
+    assert insulation.equals(insulation_pooled3)
+
+    # IV. Insulation with string or float inputs for window sizes should work.
+    calculate_insulation_score(clr, '10_000_000')
+    calculate_insulation_score(clr, '10_000_000', nproc=2)
+    calculate_insulation_score(clr, '10_000_000', nproc=3)
```
gfudenberg, line 218: similar comment-- can refactor to just keep the insulation.equals(insulation_pooled2) inside of existing earlier test to minmize the boilerplate code

gfudenberg:
I think the new tests can be made more concise-- e.g. for each tool can just refactor to keep the insulation.equals(insulation_pooled2) inside of existing earlier test to minmize the boilerplate code

gfudenberg:
note this would close #431

tests/test_snipping.py
Yaoyx: That makes sense. I will only keep one

gfudenberg:
hi @sergpolly & @nvictus -- could either of you check out this PR? Thanks!

cooltools/lib/common.py
nvictus, line 1: You might want to decorate the inner wrapper function in the `pool_decorator` itself with the [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps) decorator

cooltools/api/dotfinder.py
sergpolly: I left that comment there to try `cooler.tools.MultiplexDataPipe` in the future for describing the dots-pipeline. I think this `DataPipe` is used currently in the coverage calculations...

Unless @nvictus plans to retire `cooler.tools.MultiplexDataPipe` , I would leave that comment there - maybe change it to a more obvious: `# TODO: try using cooler.tools.MultiplexDataPipe for pipelining the steps`

cooltools/lib/common.py
```diff
@@ -532,15 +532,14 @@ def wrapper(*args, **kwargs):
                 mymap = pool.imap
             else:
                 mymap = map
-
-            func(*args, **kwargs, map=mymap)
-
+            try:
+                result = func(*args, **kwargs, map=mymap)
```
sergpolly, line 536: so, `pool_decorator` can decorate function that both have `nproc` and `map` as key word arguments, correct ?

cooltools/lib/common.py
```diff
@@ -532,15 +532,14 @@ def wrapper(*args, **kwargs):
...
+            finally:
+                if pool != None:
```
sergpolly, line 538: this should guarantee the destruction of the pool (and killing children-processes) in case `func` crashes or doesn't have `map` as an argument - right ?

cooltools/api/dotfinder.py
```diff
@@ -1307,20 +1309,16 @@ def scoring_and_histogramming_step(
     # standard multiprocessing implementation
     if nproc > 1:
```
sergpolly, line 1310: perhaps such a logging message should be hidden in the decorator itself ?

cooltools/api/dotfinder.py
```diff
@@ -1307,20 +1309,16 @@ def scoring_and_histogramming_step(
...
         map_kwargs = {}
-    try:
-        # consider using
-        # https://github.com/mirnylab/cooler/blob/9e72ee202b0ac6f9d93fd2444d6f94c524962769/cooler/tools.py#L59
-        histogram_chunks = map_(job, tiles, **map_kwargs)
-    finally:
-        if nproc > 1:
-            pool.close()
+        
+    map_ = map
```
sergpolly, line 1317: this looks confusing to me - perhaps we can rename the name corresponding kwarg in the function itself from `map` to `_map` or `map_` or `local_map` and then just apply it as is without such a renaming `_chunks = local_map(job, tiles, **map_kwargs)`

cooltools/lib/common.py
```diff
@@ -532,15 +532,14 @@ def wrapper(*args, **kwargs):
...
+            return result 
         else:
-            raise TypeError(
-                "nproc must be specified as a keyword argument (e.g. nproc=1, nproc=5...)"
-            )
-
-        if pool != None:
-            pool.close()
-
+            warnings.warn("nproc is not specified, single thread is used")
```
sergpolly, line 542: I wonder if logging is more appropriate here ? Also maybe it would be worth logging about creating a pool of nproc workers when nproc>1 as well ?

sergpolly (changes requested):
I think this looks good, but maybe it is worth hiding the logging part inside the decorator itself to announce serial processing vs pool of nproc workers and also get rid of that renaming step inside decorated functions `map_ = map`, by renaming the name of the map argument itself

cooltools/lib/common.py
sergpolly, line 1: one more thing - I think it would be nice to improve the docstring of the `pool_decorator`, specifically explicitly explain there what kind of functions can it decorate and how:
```python
"""
A decorator function that enables multiprocessing for a given function.
Function must have nproc and local_map as key word arguments.
It works by hijacking locasl_map argument and substituting it with the
parallel one: multiprocess.Pool(nproc).imap, when nproc > 1

func : callable
   A function to be decorated. Must receive nproc and local_map as key word arguments

...
"""
```

cooltools/api/dotfinder.py
nvictus: `cooler.tools  is deprecated in favor of `cooler.parallel`

cooltools/lib/common.py
nvictus, line 538:
```suggestion
                if pool is not None:
```

cooltools/lib/common.py
nvictus: Yes, this should be documented.

cooltools/lib/common.py
nvictus: [Docs](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool) seem to say that calling `terminate()` is also needed:

image

cooltools/api/dotfinder.py
nvictus: why does it need renaming at all?

nvictus: The `map` parameter should be shadowing the builtin map inside the function. If the builtin version is really needed, all you need to do is `import builtins` and use `builtins.map`.

If this is considered too ugly, then I think the alternative should be an unambiguous name for the function parameter, i.e. what should the `@pool_decorator`-compatible function protocol be: `nproc` + ??? I might suggest something like `map_func` or `map_functor`.

nvictus (changes requested):
The pool decorator is a great idea and this overall looks good to me! Just one important thing missing.

As it stands right now, applying `@pool_decorator` to a function removes the flexibility that the `map=` parameter provides in the first place, which is to support alternative and third party map functors (e.g. from `concurrent.futures` or `loky`). This decorator forces the function to always use `multiprocess.Pool.imap` or `builtins.map` when `nproc` is 1.

The decorator should provide a sane default, like `imap` (or could even be parameterized in the future `@pool_decorator("imap_unordered")`). But it should allow the client to explicitly pass in `map=<my_custom_functor>` which will override the defaults and ignore `nproc`.

cooltools/api/dotfinder.py
sergpolly: ok , than this is comment about `cooler.parallel.MultiplexDataPipe`, bit is still relevant - would be nice to replace to:
```python
# TODO: try using cooler.parallel.MultiplexDataPipe for pipelining the steps
```

cooltools/api/dotfinder.py
sergpolly: +1 for `map_functor` - this make an intent clear, i think - i.e. `map_functor` is a `map` that has been messed with. And get rid of that `map_ = map` inside

gfudenberg:
these changes look great @Yaoyx !

two qs:
- we currently let the user pass their own map to map_functor when nproc=1. However, this map could use more than one process. Would it be more clear to require nproc=None for the case when the user wants to pass their own map function ? cc @nvictus @sergpolly 

e.g.
```python
@pool_decorator
def cooltools_function_to_decorate(
    clr, ...
    nproc=None,
    map_functor=my_custom_map_func,
):
```


- seems that the user-facing `dots()` function could be modified to take the map_functor & the inner scoring_and_histogramming etc could take the default nproc=1 as the kwarg? @Yaoyx

sergpolly:
> * Would it be more clear to require nproc=None for the case when the user wants to pass their own map function

... no strong opinion ... I think, good **docs** would be sufficient - sort of like - "as soon as you start messing with the `map_functor` you - the user - are on your own - i.e reponsible for all of the boiler plate code around starting and terminating processing pool and the `nproc` doesn't affect anything in this case" Or in other words - "if you (the user) want an "easy" multiprocessing -> just modify `nproc` cooltools/decorator will take of the rest, otherwise if you want something custom - modify `map_functor` and ignore `nproc`.

> seems that the user-facing dots() function could be modified to take the map_functor & the inner scoring_and_histogramming etc could take the default nproc=1 as the kwarg

in the same vein - I'd say, _ultimately_, provide the convenience of `nproc` only for the user-facing functions .. Inner functions can receive `map_functor` argument only. Unless I'm missing something ... Also, I think `dots` is the only place where this would matter for now - i.e. the case where multiple inner functions are using parallel-`map_functor`

sergpolly:
oh , i just looked at the overall code - and there is 1 more tiny thing - I think it would be also nice to clean up all the remaining `import multiprocess/int` statements from the modules/files the rely on `pool_decorator` now - perhaps even black/ruff code formatter would take care of that - i.e. the imports that aren't being used inside functions

gfudenberg (commented):
let's try to have the same defnition for the nproc & map_functor arguments across functions. I propose:

nproc : int, optional
        How many processes to use for calculation. Ignored if map_functor is passed.

    map_functor : callable, optional
        Map function to dispatch the matrix chunks to workers. If left unspecified, pool_decorator applies the following defaults: if nproc>1 this defaults to multiprocess.Pool; If nproc=1 this defaults  the builtin map. 

does this clear @sergpolly ?

sergpolly:
yes, I think, it is clear

The last thing on my side, @Yaoyx - i've noticed few remaining `import multiprocess` in some CLI files - "sample" and "coverage" at least I didn't check them all -> otherwise this looks complete - great stuff !

sergpolly (approved):

gfudenberg:
ok to merge now @nvictus ?

cooltools/lib/common.py
```diff
@@ -512,7 +507,10 @@ def align_track_with_cooler(
 
 def pool_decorator(func):
     """
-    A decorator function that enables multiprocessing for a given function.
+    A decorator function that enables multiprocessing for a given function.  
+    Function must have nproc and map_functor as key word arguments.
```
nvictus, line 511: The wrapped function doesn't actually need `nproc`, does it? The wrapper takes it in and uses it to initialize a functor.

nvictus (approved):
Other than the docstring comment, this LGTM. One issue is that the `nproc` parameters lingering in the wrapped functions aren't necessary because `nproc` is provided by the decorator and doesn't get passed down. We can consider removing them in the future, but maybe retaining them in docstrings.

cooltools/lib/common.py
nvictus, line 511:
```suggestion
    The function must have a ``map_functor`` keyword argument.  
```

sergpolly: do I understand correctly , that when wrapped function receieves a bunch of `kwargs` one can choose to pass only a subset of those down to the `func` being wrapped ?

So you're suggesting `nproc` should be `kwargs.pop()` inside the decorator , right ? (eventually - not right now)

nvictus: Well, yes, if you look, `nproc` doesn't get used in any of the wrapped functions at all. There is no need to pass it down since it is just used to initialize a map functor. From the client's perspective, however, the decorated function does accept a `nproc` argument.

gfudenberg:
how would you then suggest CLI compatibility @nvictus ? could the map_functor argument be an int in which case the decorator interprets it as `nprop=my_int`?

nvictus:
Yeah, assuming the CLI is not customizing the map implementation, you only need to expose a parameter to set the number of processes.

gfudenberg:
I like it! That would make the arguments a little cleaner by not having 2 arguments to control 1 behavior... Though providing an int to an argument named map_functor could be a bit confusing...

cooltools/lib/common.py
sergpolly: I see I see - took me a while to realize the kwargs go to the wrapper first and the wrapped func doesn't have to "see" all of the kwargs ...

Now the docstring interplay is unclear to me - i.e. you'd need to document nproc in the wrapper then and somehow combine docstrings of the wrapper and the wrapped func ?

sergpolly:
@gfudenberg a function wrapped into pool_decorator - has both nproc and map-functor, but the function being wrapped does not have to have nproc as a kwargs, only map_functor - I think this was Nezar point ... At least this is how understand it now - took time to understand the interplay of the wrapper and the wrapped

gfudenberg:
yes, this is now my understanding too-- so I'm wondering if 

```
@pool_decorator
def cooltools_function_to_decorate(
    clr, 
    ... other args,
    map_functor=1 or int or my_custom_map_func(),
):
```
would be clear, or if this would make the map_functor argument confusing (or better to rename it something else)

sergpolly:
Hm , I don't know if passing map_functor=8 or int would improve readability... IMHO. I very much like the current state + Nezars suggestion - the way I understand it is
- make cooltools_func receive only map_functor  kwarg
- wrapping cooltool_func in a pool_decorator makes it seem like the function has nproc kwarg now (pool decorator - organizes the interplay between nproc and map_functor internally)
- inside cli we use decorated cooltool_func with nproc only without exposing/touching map_functor

The only part missing for me now - is how do you organize docstrings... - cooltools_func drops nproc as kwarg , together with the docstrings for it - ok, and then what ? Do we specify docstring for nproc in the wrapper ? Can one combine the docstrings of wrapper the wrapped up in a nice way ?

sergpolly:
@gfudenberg I'm still a bit confused about this nproc/map_functor thing ... am I missing something obvious ? or do you agree with the previous comment ?

gfudenberg:
sorry for the confusion @sergpolly ! 
your comment makes perfect sense & I agree with the following:

> The only part missing for me now - is how do you organize docstrings... - cooltools_func drops nproc as kwarg , together with the docstrings for it - ok, and then what ? Do we specify docstring for nproc in the wrapper ? Can one combine the docstrings of wrapper the wrapped up in a nice way ?

sergpolly:
perfect ! it is nice to have a consensus ! i'm sure there is some python/decorator/wrapping magic that would enable docstring combination

PR #493, opened 2024-01-29
Update _numutils.pyx

vishalt12345:
Changed "random" to "rand" -- The random function was not available on MSVC's "stdlib.h" for me, and after doing some more research, I realized that "rand()" instead of "random()" might be better for more compatibility across various operating systems.

### DIFF
--- a/cooltools/lib/_numutils.pyx
+++ b/cooltools/lib/_numutils.pyx
@@ -13,7 +13,7 @@ ctypedef unsigned short ushort
 ctypedef unsigned char uchar
 
 cdef extern from "stdlib.h":
-    long c_libc_random "random"()
+    long c_libc_random "rand"()
     double c_libc_drandom "drand48"()
 
 


sergpolly:
hm , here https://stackoverflow.com/questions/76663458/do-rand-and-random-function-the-same-in-glibc they recommend using `rand_s` on windows ...

Should we try a system specific macro instead ?
i'm not that familiar with Cython unfortunately

sergpolly:
another suggestion/proposal - rewrite `fake_cis` using `numba` - I know it is much more than just changing import statement, but nonetheless - would you be interesting in trying this out ?

That `rand` function you mentioned is used only in one function inside the `_numuitls.pyx` module - which is `fake_cis` function - it doesn't seem very complicated - it received 2 2D arrays, one with data and another one with the boolean mask (typically data would be entire hic heatmap and 2D mask would be `True` for the intra-chromosomal interactions). Function simply replaces data in the masked region (in cis region) with the random value from the same row - hence the name `fake_cis` - this is done to perform eigendecomposition on entire HiC heatmap, such that cis-data does not "interfere" with trans data

we started discussing simplifying cython dependencies in favor of numba here https://github.com/open2c/cooltools/discussions/487

PR #513, opened 2024-04-12
Allow upgrade pandas (version 2+) and python (up to 3.12) versions

thomas-reimonn:
This PR adds cool tools support for updated pandas versions and python versions.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/pytest.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ '3.8', '3.9', '3.10' ]
+        python-version: [ '3.9', '3.10', '3.12' ]
```
mimakaev, line 18: why not 3.11?

mimakaev (approved):
Looks good to me

PR #515, opened 2024-04-20
update version and changelog

Yaoyx:
Update version to 0.6.2

### DIFF
--- a/cooltools/__init__.py
+++ b/cooltools/__init__.py
@@ -11,7 +11,7 @@ The tools for your .cool's.
 """
 import logging
 
-__version__ = "0.6.1"
+__version__ = "0.6.2"
 
 from . import lib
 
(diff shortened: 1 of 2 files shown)


cooltools/__init__.py
```diff
@@ -11,7 +11,7 @@
 """
 import logging
 
-__version__ = "0.6.1"
+__version__ = "0.6.2"
```
nvictus, line 14: If there are public API changes and/or new features, you want to do a minor bump. Patch bumps are for bug fixes. See [semantic versioning](https://semver.org/). 

Note that we are still in beta (major 0), we don't worry about major version bumps until our API is stable.

```suggestion
__version__ = "0.7.0"
```

Yaoyx: Ah ha, thank you! I update it to 0.7.0.

PR #531, opened 2024-08-09
reduce the chunksize in rearrange to a more reasonable 2e7

golobor:


golobor:
Hi, @Phlya , did we agree on Slack if this is a good idea?

Phlya:
I remember we talked about it... Was it taking too much memory with 1e9? I must have tested on our big server and didn't notice. If it helps, sure, sounds like a good idea :)

Phlya (approved):

Issue #520, opened 2024-05-28
`Matplotlib` 3.9 broken dependency


nvaulin:
Hi, I found that in new version of `matplotlib` (3.9) there are some functions being removed. 

For instance, for 

```python
import cooltools.lib.plotting
```
 I've got:
```python
----> 5 from matplotlib.cm import register_cmap
      6 import matplotlib as mpl
      7 import matplotlib.pyplot as plt

ImportError: cannot import name 'register_cmap' from 'matplotlib.cm' (/home/nvaulin/tools/mambaforge/envs/hic-env/lib/python3.10/site-packages/matplotlib/cm.py)
```

This function was present in 3.8, but not in 3.9 (https://matplotlib.org/3.8.4/api/cm_api.html#matplotlib.cm.register_cmap
)

So maybe you could add specific version of `matplotlib` to the `requirements.txt`, or adjust the code

Phlya:
New function to use: https://matplotlib.org/stable/api/cm_api.html#matplotlib.cm.ColormapRegistry.register

nvaulin:
> https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.9.0.html#removals
> New function to use: https://matplotlib.org/stable/api/cm_api.html#matplotlib.cm.ColormapRegistry.register

Yes, I mean, this removed func you mentioned is used In 'cooltools.lib.plotting' module.

golobor:
must be finally solved by #533. Feel free to reopen if the issue appears again!

PR #533, opened 2024-08-09
an attempt to fix register_cmap import in plotting

golobor:
addresses #520

### DIFF
--- a/cooltools/lib/plotting.py
+++ b/cooltools/lib/plotting.py
@@ -5,7 +5,8 @@ Migrated from :mod:`mirnylib.plotting`.
 try:
     from matplotlib.cm import register_cmap
 except ImportError:
-    from matplotlib.colormaps import register
+    from matplotlib import colormaps
+    register_cmap = colormaps.register
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


golobor:
@Phlya , i've requested your review, b/c you seem to have fixed #520 before (https://github.com/open2c/cooltools/commit/dc1fdd91eb4575ca56c2aae91d9e05dd18328082) , but the fix didn't work for me. Did your fix work for you? 

I think the reason your version doesn't work for me is b/c matplotlib.colormaps is not a module, but a variable.
Here is where it gets imported into the matplotlib namespace from mpl.cm:
https://github.com/matplotlib/matplotlib/blob/857e7d8aacd52d552e2e617ce23da30166419dcd/lib/matplotlib/__init__.py#L1515
And here is where it gets defined:
https://github.com/matplotlib/matplotlib/blob/857e7d8aacd52d552e2e617ce23da30166419dcd/lib/matplotlib/cm.py#L238

In any case, the new version should be agnostic to whether mpl.colormaps is a module or a variable. Would you green-light it?

Phlya:
Did they also swap the argument order? WTF. OK, looks good to me, thanks for fixing it again!

Phlya (approved):

golobor:
ah, good catch re: argument order. I will use kwargs to make sure that it's backward compatible.

Issue #556, opened 2025-07-05
`rearrange.py` should respect count datatype of source cooler


dmalzl:
I recently posted [an issue](https://github.com/open2c/cooltools/issues/554) about an Exception unnecessarily raised when chromosomes are not sorted in a specific way (though I admit sorting is not that hard). There, I was advised to use [`rearrange.py`](https://github.com/open2c/cooltools/blob/master/cooltools/cli/rearrange.py) to fix the problem without much recomputation. However, when using this I found that half of our data goes missing. A short investigation revealed that this is due to an implicit float to int casting when creating the cooler. This happens because `rearrange.py` does not account for input datatype which needs to be passed to `cooler.create_cooler` explicitly via the `dtypes` argument. If this is `None` cooler simply uses the default pixel datatype which is `int32`. Casting floats < 0 to int results in clipping them to 0 which in turn removes part of our data from the file (part of our data is floats < 0).
I suggest to determine the input pixel datatype and pass it to `cooler.create` explicitly

Phlya:
Sounds like a very reasonable suggestion, would you be able to make a PR?

dmalzl:
Sure. Will do it on Monday :)

dmalzl:
Closing after fix

PR #557, opened 2025-07-07
respect dtypes of input file

dmalzl:
This is a fix for #556

### DIFF
--- a/cooltools/api/rearrange.py
+++ b/cooltools/api/rearrange.py
@@ -202,6 +202,8 @@ def rearrange_cooler(
         bins_old, view_df, new_chrom_col=new_chrom_col, orientation_col=orientation_col
     )
     logging.info("Creating a new cooler")
+    # fetching pixel dtypes to avoid implicit casts of count column
+    dtypes = clr.pixels().dtypes.to_dict()
     cooler.create_cooler(
         out_cooler,
         bins_new,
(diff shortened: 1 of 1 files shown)


dmalzl:
Yes. I'll try :)

Phlya:
It looks good, but the test actually fails... Does it work for you locally?

dmalzl:
Yes, also just noticed. Was a premature push. i did not think about the rearranged pixels after calling `rearrange_cooler`. Currently trying to figure out if I can just bypass the order check in some elegant way.

dmalzl:
Fixed it by simply leaving the chromosome order as is and just checking if dtypes are equal. This now passes locally and I also fails when dtypes are not equal

tests/test_rearrange_cooler.py
```diff
@@ -136,3 +136,15 @@ def test_rearrange_cooler(request):
         ::-1, :
     ]
     assert np.array_equal(old_trans_m, reordered_inverted_trans_m, equal_nan=True)
+
+    # IV.
+    # Check that pixel datatypes are propagted to output cooler
```
Phlya, line 141: propagAted

dmalzl:
Thanks for the review and sorry for the inability to type straight^^

dmalzl:
Great :) Always happy to help making the ecosystem better ;). Have a nice day

